Issue #2, opened 2016-12-09
Can't get it to work installed globally


leewillis77:
I have WordPress-VIP installed globally, and PHPCS can find it just fine:

```
$ phpcs -i
The installed coding standards are MySource, PEAR, PHPCS, PSR1, PSR2, Squiz, Zend, WordPress, WordPress-Core, WordPress-Docs, WordPress-Extra and WordPress-VIP
```

If I then install this repo globally, e.g. 

```
$ composer global require humanmade/coding-standards
[snipped]
$ phpcs --config-set installed_paths /Users/leewillis/.composer/vendor/wp-coding-standards/wpcs/,/Users/leewillis/.composer/vendor/humanmade/coding-standards/
Config value "installed_paths" added successfully
```

HM shows up as an available standard:
```
$ phpcs -i
The installed coding standards are MySource, PEAR, PHPCS, PSR1, PSR2, Squiz, Zend, WordPress, WordPress-Core, WordPress-Docs, WordPress-Extra, WordPress-VIP and HM
```

However - it's unusable:

```
$ phpcs --standard=HM foo.php 

Fatal error: Uncaught PHP_CodeSniffer_Exception: Referenced sniff "WordPress-VIP" does not exist in /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer.php:1167
Stack trace:
#0 /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer.php(780): PHP_CodeSniffer->_expandRulesetReference(Object(SimpleXMLElement), '/Users/leewilli...', 0)
#1 /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer.php(578): PHP_CodeSniffer->processRuleset('/Users/leewilli...')
#2 /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer/CLI.php(956): PHP_CodeSniffer->initStandard(Array, Array, Array)
#3 /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer/CLI.php(113): PHP_CodeSniffer_CLI->process()
#4 /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/scripts/phpcs(25): PHP_CodeSniffer_CLI->runphpcs()
#5 {main}
  thrown in /Users/leewillis/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer.php on line 1167
```

[...]

PR #7, opened 2017-05-04
incorrectly detects `function() use () {}`

roborourke:
If you have a nested or lambda function using the `use` keyword this sniff picks it up and flags it as an issue.

In the above scenario the token will have `nested_parenthesis` set so this PR just adds a check for that and skips the token.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     docker:
       - image: php:7.1
-    working_directory: ~
+    working_directory: /tmp/hm-cs
     steps:
       - checkout
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,11 @@ jobs:
       - image: php:7.1
     working_directory: /tmp/hm-cs
     steps:
+      - run:
+          name: Install Git
+          command: |
+            apt-get update
+            apt-get install -y git-all
       - checkout
       - run:
           name: Install Composer
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,30 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: php:7.1
+    working_directory: /tmp
+    steps:
+      - checkout
+      - run:
+          name: Install Composer
+          command: |
+            php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
+            php -r "copy('https://composer.github.io/installer.sig', 'composer-setup.sig');" && \
+            php -r "if (hash_file('SHA384', 'composer-setup.php') === trim(file_get_contents('composer-setup.sig'))) { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" && \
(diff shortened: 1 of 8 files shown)


grappler:
It may be good to add a test for it too.

roborourke:
Didn't spot the test suite 😳

roborourke:
@grappler so I don't think any of this is tested at the moment, not sure how to do it for code sniffer libraries yet

grappler:
Ah, yes. :(

It is not that difficult. The folder structure is 
- HM
  - Tests
    - Layout
      - OrderUnitTest.inc (This contains the test code)
      - OrderUnitTest.php (This tells the test in which line the code is bad.) You can see an example here. https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/Arrays/ArrayKeySpacingRestrictionsUnitTest.php

As there is no Travis set up you would need to manually test it. Documentation https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/CONTRIBUTING.md#unit-testing

roborourke:
@grappler I hit a snag when I was trying to do this yesterday, code sniffer is required downstream as a dependency however the tests directory isn't part of the package. How do you get around it on wpcs?

grappler:
I am not sure I understood you correctly.  This is how you would install phpunit in phpcs.

```
which phpcs
cd /path/to/phpcs
composer install
vendor/bin/phpunit --filter WordPress tests/AllTests.php
```

If you have phpunit installed elsewhere then you could just run
`phpunit --filter WordPress /path/to/phpcs/tests/AllTests.php`

roborourke:
@grappler ohhh, the `cd` into phpcs bit was what I was missing. Let me try that

roborourke:
Also I need a clone of phpcs right?

grappler:
You should have phpcs on your system otherwise the sniffing would not work. It depends on how you have installed it. It will not work if you have installed it as phar.

roborourke:
I think I installed via brew originally. I've cloned the repo and set up my path to use that but currently it doesn't seem to be saving the installed paths in it's config :/

grappler:
I think the reason is due to how the paths are defined in the `ruleset` https://github.com/humanmade/coding-standards/issues/2

roborourke:
@rmccue can you take a gander at this when you get some time? Need to test the paths still work when using it in a project actually. Anyway, have a play with the tests, I added docs on running them to the readme.

- [ ] TODO: create a docker image with php, git, composer & phpcs already checked out and this will super fast

rmccue:
@roborourke FWIW, I'd rather see the test suite in a separate PR (although you can add the test cases in this one if you want).

Not sure if Circle supports it, but Travis supports caching Composer dependencies and stuff pretty easily. (I'd also rather we use Travis than Circle for open source stuff generally, but I don't care that much, and you've already written it.)

If you want to split the unrelated changes into a separate PR, happy to merge the `use` change straight away.

roborourke:
@rmccue that's interesting, I thought we were settling on circle as a company! I'll make a new PR for the `use` changes so it's clean.

roborourke:
Closed in favour of #8 and #9

PR #8, opened 2017-05-09
Fix order sniff, add tests

roborourke:
The order sniff was incorrectly picking up the use keyword in nested lambda functions. This fixes it and adds a unit test for the sniff.

cc @rmccue


rmccue:
Thanks @roborourke.

We should probably also test the current behaviour to ensure this doesn't break it. I'll file a new issue for that.

roborourke:
Testing wise this is one is a weird case - it always reports an error on
line one and only reports one error even if more things are out of order. I
couldn't really figure it out.

Would need to improve the way the sniff returns it's errors.

PR #12, opened 2017-05-25
Fix false positives for T_USE (Fix #5)

### LINKED ISSUE #5: HM.Layout.Order.WrongOrder false positives
`HM.Layout.Order.WrongOrder` can throw a false positives in files with anonymous functions using use.

This code 

```php
<?php
namespace Hamilton;

use Hamilton\Mixtape;

const WRITER = 'Lin-Manuel Miranda';

add_filter( 'Hamilton/Cast', function ( $cast ) use ( $replacements ) {
	return $replacements;
} );
```

Error thrown:

```
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 1 | ERROR | use found on line 8, but const was declared on line 6.
   |       | Statements should be ordered `namespace`, `use`, `const`,
   |       | `require`, then code. (HM.Layout.Order.WrongOrder)
----------------------------------------------------------------------
```

connerbw:
Test case (should only fail on last line):

```php
<?php

use Foo\Bar;
use Foo\Baz as Zztop;

require( 'some/file/that/exists.php' );

$foo = 1;

$a = function () use ( $foo ) {
	return $foo + 1;
};

$b = function ( $i, $j ) use ( $foo ) {
	return $foo + 1;
};

$c = function ( $i, $j = null ) use ( $foo ) {
	return $foo + 1;
};

$c = function ( ...$i ) use ( $foo ) {
	return $foo + 1;
};

$c = function () use ( $foo ) : int {
	return $foo + 1;
};

$c = function ( ...$i ) use ( $foo ) {
	return $foo + 1;
};

$c = function ( ...$i ) use ( $foo ) : int {
	return $foo + 1;
};

$c = function ( ...$i )

use ( $foo ) : int {

	return $foo + 1;

};

add_action( 'init', function () use ( $foo ) {
	return $foo + 1;
}, 1000 );

$c = function() {
	return true;
};

use Hello\World;

```


joehoyle:
ping @rmccue

PR #13, opened 2017-05-25
Fix rule when runing on Windows

connerbw:
`DIRECTORY_SEPARATOR` is equal to `\` on Windows, but the code uses a hard coded `/inc` to do its checks.

This fix normalizes `$directory` when running on Windows.

### DIFF
--- a/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
+++ b/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
@@ -43,12 +43,12 @@ class NamespaceDirectoryNameSniff implements PHP_CodeSniffer_Sniff {
 
 		$full = $phpcsFile->getFileName();
 		$filename = basename( $full );
-		$directory = dirname( $full );			 
+		$directory = dirname( $full );
 
 		// Normalize the directory seperator accross operating systems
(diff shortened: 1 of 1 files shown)


HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
```diff
@@ -44,6 +44,11 @@ public function process( PHP_CodeSniffer_File $phpcsFile, $stackPtr ) {
 		$full = $phpcsFile->getFileName();
 		$filename = basename( $full );
 		$directory = dirname( $full );
+		
+		// Normalize the directory seperator accross operating systems
+		if ( strtoupper( substr( PHP_OS, 0, 3 ) ) === 'WIN' ) {
```
rmccue: Can we make this `DIRECTORY_SEPARATOR !== '/'` and `str_replace( DIRECTORY_SEPARATOR, '/', $directory )` please? :)

connerbw: Great idea. Will fix.

rmccue:
Thanks! You had some trailing whitespace in the patch, so I cleaned that up. :)

PR #25, opened 2017-11-13
Use the short array checker from PHPCS

joehoyle:
This gives us the fixer for free.


joehoyle:
@rmccue wondering if there's a reason you didn't use this in the first place?

Issue #27, opened 2017-11-17
Fix unit tests on this repo


rmccue:
We should have real unit tests here.

PR #30, opened 2017-11-21
Exclude space checks from JS files

rmccue:
This avoids the JS-in-PHP parser thinking that `<Component>` is actually comparison operators and complaining endlessly about it.


joehoyle:
@rmccue approved, but not sure if this failure is new?

rmccue:
Nah, unit tests on this repo are broken, see #27.

PR #31, opened 2017-11-22
Update phpcs to v3

rmccue:
This gets us a bunch o' new stuff, including:

* Parallel checks with `--parallel=10`
* Caching between runs with `--cache`
* `--report=code` to show the errored line in context
* `-m` to suppress error message gathering to save memory
* Autodetection of phpcs files
* ES6 tokenisation
* Progress with `-p`

[All changes](https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0).

### DIFF
--- a/HM/Sniffs/Classes/OnlyClassInFileSniff.php
+++ b/HM/Sniffs/Classes/OnlyClassInFileSniff.php
@@ -2,14 +2,13 @@
 
 namespace HM\Sniffs\Classes;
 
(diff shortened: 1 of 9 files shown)


joehoyle:
Will everyone need to upgrade their copy of PHPCS too?

rmccue:
> Will everyone need to upgrade their copy of PHPCS too?

Not until I release the next version in any case.

PR #36, opened 2018-01-30
Only run PHPCS on PHP files.

joehoyle:
Running css and js thorugh phpcs is causing too many complications, we should use eslint directly for JS and do _something_ with css too I guess.


mikeselander:
> do something with css too I guess.

There's some stylelinting tools that @peterwilsoncc's  going to write up for the CSS Standards so we're covered on that.

rmccue:
I'd rather we leave the sniff there, but just disable it, and write up how to enable it in the README. Alternatively, phpcs itself includes an ESLint sniff in its core now, so we could document how to use that.

joehoyle:
@rmccue updated to just remove JS / CSS files from the default. However, I don't know why we'd document ESLint via PHPcs when it's pretty broken. We should add ESLint support as it's on thing, as a separate repo(maybe).

rmccue:
@joehoyle Running ESLint via phpcs isn't *that* broken, it's more that phpcs' JS tokeniser is horrifically broken.

PR #39, opened 2018-03-06
Add support for a .phpcsignore file

rmccue:
To match `.eslintignore` and `.gitignore`, we should support a `.phpcsignore`. This removes one of the big reasons you'd need a `phpcs.ruleset.xml` in many repos.

See https://github.com/squizlabs/PHP_CodeSniffer/issues/1884

### DIFF
--- a/HM/bootstrap.php
+++ b/HM/bootstrap.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace HM\Sniffs;
+
+/**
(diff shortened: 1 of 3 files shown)


joehoyle:
@rmccue we need anythign else here?

rmccue:
I think we're good, we should deploy to linter-bot

PR #40, opened 2018-03-16
Move YAML rule definitions into index.js

kadamwhite:
This PR moves our rule definitions from a YAML file into index.js itself.

ESLint rule configuration examples are broadly provided in JSON format, not YAML. Moreover, while one generally alredy knows JSON if familiar with YAML, the opposite is not always true. Rather than requiring an additional parsing library and making the barrier to entry for potential contributors higher, I'm proposing we use the JS file exports for our rule definitions instead.

I considered moving to JSON, but the parity of JSON and JS formatting and the simplicity of having our JS rules defined in JS (where we can comment, process and format as normal) felt best. The `.eslintrc` in the project now just points back to index.js.

The rules should not have been altered in their transition from YAML to JS, _with the exception_ of jsx-curly-spacing where the `children` rule had to be removed because it is not available in the declared peerDep version of eslint-plugin-react. I will reinstate it in a dependency upgrade PR.


packages/eslint-config-humanmade/index.js
```diff
@@ -1,6 +1,87 @@
-const fs = require("fs");
-const path = require("path");
-const yaml = require("js-yaml");
-
-const configPath = path.join(__dirname, ".eslintrc.yml");
-module.exports = yaml.safeLoad(fs.readFileSync(configPath, "utf8"));
+/* eslint quote-props: ["error", "always"] */
+// Quote properties for consistency in this file only
+// (Many ESLint rules use key names that require quotation marks)
```
kadamwhite: Note that this file does not yet pass itself, but that's a step for a follow-up PR.

kadamwhite: Actually upon reflection we don't define any `quote-props` rule so this comment is superfluous, but I've removed it in a follow-up.

kadamwhite:
Thanks!

PR #41, opened 2018-03-19
Make index pass its own rules

kadamwhite:
This upgrades eslint, adds the “consistent” option for single property object sanity, and adjusts the rules and config file so that it passes its own definitions.


rmccue:
Can we also add some test files (maybe in a new PR) to ensure that we're not causing regressions?

kadamwhite:
Indeed we can! I’ve begun that work locally, it will be a follow up PR today.

PR #42, opened 2018-03-19
Add ESLint config test script with example fixtures

kadamwhite:
I started going down a dark path trying to use Jest to verbosely specify all of our style configuration options, but that felt like it was duplicating the work ESLint's own test suite does to no clear gain.

Since the intent is for us to be able to validate that our own rules pass on a set of files that look representative of our own standards, this PR takes the alternative approach of adding a single node script that will run ESLint against a directory of expected-to-fail antipatterns, and a directory of expected-to-pass exemplary code files.

If there are no errors where there should be errors, or if there are errors where there should not be, the script will exit with an error.


kadamwhite:
@rmccue Could you let me know the status of this PR?

packages/eslint-config-humanmade/fixtures/antipatterns/component-jsx-perentheses.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
rmccue: Filename is misspelled, should be `parentheses`

packages/eslint-config-humanmade/fixtures/test-lint-config.js
```diff
@@ -0,0 +1,61 @@
...
+	warnings: counts.warnings + file.warningCount,
+	files: counts.files + 1,
+} ), {
+	errors: 0,
+	warnings: 0,
+	files: 0,
+} );
+
+console.log( 'Running ESLint on fixture directories. Use --verbose for a detailed report.' );
+console.log( '\nLinting `fixtures/antipatterns/**`...' );
```
rmccue: I'd rename the directory to `fixtures/fail` to make it clearer what is happening.

packages/eslint-config-humanmade/fixtures/test-lint-config.js
```diff
@@ -0,0 +1,61 @@
...
+
+const antipatternReport = cli.executeOnFiles( [ join( __dirname, 'antipatterns/**' ) ] );
+const antipatternCounts = count( antipatternReport.results );
```
rmccue: This should go file-by-file, otherwise having any file that fails in the directory will be fine, even if there's a regression in another place.

packages/eslint-config-humanmade/index.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
 			'consistent': true,
 		} ],
 		'object-curly-spacing': [ 'error', 'always' ],
+		'object-property-newline': [ 'error' ],
```
rmccue: Does this need the one-property exception, or is that a different rule?

rmccue:
@kadamwhite I've updated per above feedback, and it's now running via Travis as well. Can you review my changes please? :)

rmccue:
@kadamwhite Any thoughts? 😬

packages/eslint-config-humanmade/index.js
kadamwhite: I _think_ that this does not need the one-property exception; but it will error on e.g.
```
const obj = { a: 'multi', property: 'object' };
```
so if that's not desired you can remove this. Can't 100% recall why I felt it was important to add here.

kadamwhite:
@rmccue I can't add a code review to a PR I opened, but these changes look great! :shipit:

PR #48, opened 2018-04-18
Allow use of filesystem functions

joehoyle:
For I think not good reasons, the use of fopen, etc are not recommended, and WP_Filesystem is instead recommended. I don't agree with this, and recommend we should instead encourage them. The filesystem functions provide a great abstraction for reading and writing streams, and WP_Filesystem on the other hand... well yeah.


rmccue:
I literally wrote the same thing just yesterday. We can do this via configuration instead of exclusions though, I'll push a PR for it.

rmccue:
Fixed in #50 instead.

PR #49, opened 2018-04-18
WIP: Ignore array item alignment rule

### LINKED ISSUE #47: Remove array alignment rule for PHP?
In #41 we removed the object property alignment rule for JS, making this valid code:

```js
const x = {
	foo: "bar",
	longerPropertyName: "value",
};
```

Should we do the same for PHP? Currently, this code fails:
```php
$x = [
	'foo' => 'bar',
	'longerPropertyName' => 'value',
];
```

The error is:
```
Array double arrow not aligned correctly; expected 16 space(s) between "'foo'" and double arrow, but found 1.
(WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
```

rmccue:
Fixes #47. Do not merge.

### DIFF
--- a/HM/ruleset.xml
+++ b/HM/ruleset.xml
@@ -24,6 +24,9 @@
 		<exclude name="WordPress.XSS.EscapeOutput" />
 		<exclude name="WordPress.VIP.AdminBarRemoval" />
 
+		<!-- Exclude other conflicting rules -->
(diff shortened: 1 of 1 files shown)


rmccue:
This turns out to be a regression, so we should fix in 0.4.2 instead.

PR #72, opened 2018-08-06
Update README to reflect updated .eslintrc location

kadamwhite:
This updates the root-level README documentation around usage of ESLint. The existing README referenced an incorrect path to the project's `.eslintrc` file, which was removed from the root directory in #54. #54 also removed the ability to run ESLint via phpcs (it is preferable to run eslint directly via npm packages, or to use linter-bot).

This PR updates the documentation around how to reference the `.eslintrc` file (renamed from `.eslintrc.yml` in #40) if you choose to use the configuration from the composer package in lieu of installing directly via npm.

Thank you @mattheu for raising this issue

### DIFF
--- a/HM/Sniffs/Debug/ESLintSniff.php
+++ b/HM/Sniffs/Debug/ESLintSniff.php
@@ -13,7 +13,7 @@ class ESLintSniff implements Sniff {
 	/**
 	 * Path to default configuration.
 	 */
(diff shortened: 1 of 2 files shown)


kadamwhite:
Merging is blocked as there are PHPUnit failures on the `master` branch. But thanks for the review @tfrommen !

PR #82, opened 2018-08-20
Update PHP Standards.

### LINKED ISSUE #108: NoSpaceAfterCloseParenthesis is (currently) inconsistent.
The `WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis` rule is (currently) inconsistent.

I get:
```
Space between opening control structure and closing parenthesis is required
(WordPress.WhiteSpace.ControlStructureSpacing.NoSpaceAfterCloseParenthesis)
```
For:
```
function test2( array $foo ): \Generator {
    yield 1;
}
```

But no complaints for the exact same code without a function parameter:

```
function test1(): \Generator {
    yield 1;
}
```

It might be fixed upstream?

See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/547#issuecomment-444241654

tfrommen:
This PR includes:

* update WPCS to v1;
* improve/fix ruleset descriptions;
* adapt ruleset to updated standards.

Also, the ruleset has been improved:

* specify _current directory_ as **default** path (which can be overriden via command line);
* do not exlude `*.css` and `*.js` files, but really restrict PHPCS to PHP files only.

~This PR also includes #31.~

Resolves #108


ntwb:
It'd be great to land this, without it I can't update WPCS to 1.0 locally as Composer SemVer requires 0.4.2 or 0.5.0 to use WPCS 0.14.0 /shrug

composer.json
```diff
@@ -1,12 +1,12 @@
...
-	"description": "Human Made coding standards",
+	"description": "Human Made coding standards.",
 	"type": "project",
 	"license": "GPL-2.0-or-later",
 	"require": {
-		"wp-coding-standards/wpcs": "^0.14.0",
 		"fig-r/psr2r-sniffer": "^0.5.0",
-		"squizlabs/php_codesniffer": "~3.2.0"
+		"squizlabs/php_codesniffer": "~3.2.0",
+		"wp-coding-standards/wpcs": "^1.0.0"
```
mikeselander: Might as well use [1.2.1](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases) at this point.

HM/ruleset.xml
```diff
@@ -1,33 +1,25 @@
...
 	<exclude-pattern>node_modules/*</exclude-pattern>
-	<exclude-pattern>*.js</exclude-pattern>
-	<exclude-pattern>*.css</exclude-pattern>
-
-	<autoload>./bootstrap.php</autoload>
-
-	<!-- Include everything in the VIP standard... -->
-	<rule ref="WordPress-VIP">
-		<!-- ...Except for VIP-specific things -->
-		<exclude name="WordPress.VIP.FileSystemWritesDisallow" />
```
mikeselander: Instead of deleting all of these rules, we should include the VIP standards separately. This would remove a lot of expected behavior from how the standards currently behave.

tfrommen: Hm, I don't really understand what you did here, and why, @mikeselander. 🙂
Of course, I did **not** just _delete_ all these rules. As of WPCS v1, the VIP ruleset is deprecated. Internally, it was using existing rules, and most of the times customized these.
So I replaced all used VIP rules with the exact setup of rules they were referring to and adapting internally. Please see the [1.0.0 release notes](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases/tag/1.0.0) for that, as well as the (now deprecated) VIP ruleset. For example, [this block](https://github.com/humanmade/coding-standards/pull/82/commits/deb4437aee5565a2784aa31bbb585d6de358bcc3#diff-b917176d6439ff1710f344ca68150a48R77) from line 77 to 91 (on the right) is the equivalent of several deprecated VIP rules...

So, using this PR, everything should have behaved just like before.

Now, with more than four months gone by, this might have changed, yes. And also, switching to the _VIP Minimum_ ruleset is a rather different thing.

I would have liked your change better in a separate PR, branched off mine, if you wanted. 🤷‍♂️

mikeselander: You've rightly pointed out that VIP-Go has a higher likelihood of introducing changes than this PR in the previous state that it was, so I'm going to branch those changes off of this PR and we can address those changes separately and more granularly 👍

nathanielks:
This has been approved, can it be merged?

composer.json
joehoyle: Do we know to what degree things have changed in 1.0.0? I don't really follow wpcs at all, but I know there's been a lot of churn, and frankly adding of pretty arbitrary "standards", so this just makes me a bit nervous. If that's all gravy, and we feel we know what we are getting ourselves in for, no worries.

HM/ruleset.xml
```diff
@@ -51,88 +43,82 @@
...
-
-	<!-- File Naming: `name/space/` -->
-		<!-- Handled in Files/NamespaceDirectoryNameSniff -->
-
-	<!-- File Naming: `class-classname.php` -->
-		<!-- Handled in Files/ClassNameFileSniff -->
+	<rule ref="WordPress.WP.CronInterval" />
+	<rule ref="WordPress.WP.CronInterval.CronSchedulesInterval">
+		<type>error</type>
+		<message>Scheduling crons at %s sec ( less than %s minutes ) is prohibited.</message>
```
joehoyle: This is the first I'm seeing of this rule, just wondering what the thought process here is? Also, what _is_ the minimum time it's going to warn on?

tfrommen: @joehoyle this rule has previosuly been _hidden_ inside the (now deprecated) VIP standards. So it has always been there. 🙂 What I did in this PR is, amongst other things, to inline all VIP rules (that we did not override/exclude).

The minimum value is [set to 15 minutes](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/c9eaadaafefce36b3cb7e06eb15305b8c4cae9ce/WordPress/Sniffs/WP/CronIntervalSniff.php#L31-L40) (due to VIP requirements).

joehoyle: 👍 awesome

composer.json
tfrommen: @joehoyle from 0.x to 1.0.0, there did change a lot, **most** of it being re-structuring of rules, plus a couple deprecations here and there... In this PR, I pretty much made the resulting rules be equivalent to what we had before.

But, to be fair, this PR is almost half a year old, and **WPCS is already at version 2.0.0** at the moment. Just saying...

ntwb:
@joehoyle wrote:
> Do we know to what degree things have changed in 1.0.0? I don't really follow wpcs at all, but I know there's been a lot of churn, and frankly adding of pretty arbitrary "standards", so this just makes me a bit nervous. If that's all gravy, and we feel we know what we are getting ourselves in for, no worries.

FWIW, WPCS 2.0.0 has since been released:

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases/tag/2.0.0

HM/ruleset.xml
mikeselander: fwiw I think we should consider removing this rule totally but that's for another PR.

mikeselander:
I'm going to merge this and we will be dogfooding it immediately on a client project. If any updates are needed we should know quickly and before we tag 0.6.

Issue #61, opened 2018-05-05
Alignment with equals signs


joehoyle:
There's seems to be a new part of `WordPress.Arrays.MultipleStatementAlignment.NotSameWarning` that makes the following a failure:

```php
$all['read_applications'] = true;
$all['read_application'] = true;
```

Do we want this / is this intentional? This is pretty dumb I think in most situations and probably leads to be putting lots of daniel spacing everywhere so the lines are not consecutive.

rmccue:
👍 for removing this.

mikeselander:
Looks like #96 cleared this out

PR #94, opened 2018-09-27
Attempt to remove assignment alignment rule

kadamwhite:
For parity with our JS rules (which do not require alignment for objects or assignment), and as discussed in #61 and #47, we should not enforce assignment alignment.

I have likely done this incorrectly, and am not sure how best to test locally, so could use assistance getting this PR up to snuff so we can close out those issues.


kadamwhite:
@joehoyle I think yes, #96 covers what this was intended to fix.

PR #96, opened 2018-10-12
Remove alignment of equals rule

joehoyle:
See https://github.com/humanmade/coding-standards/issues/61.


rmccue:
So annoying that wpcs changed this out from under us.

PR #97, opened 2018-10-12
Allow json_encode / json_decode

joehoyle:
Given we have PHP 7+, we don't need to use wp_json_encode / wp_json_decode. Using the WP alternatives is also not easy in things like SUNRISE or wp-config.php as those functions may not be yet available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## Unreleased (0.7.0)
 
 ### Changed
- - Exclude `load.php` from `NamespaceDirectoryNameSniff`
+ - Exclude `load.php` from `NamespaceDirectoryNameSniff` #131
+ - Allow `json_encode` / `json_decode` function usage #97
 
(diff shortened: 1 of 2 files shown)


rmccue:
Hmm, I'd probably leave this one; `wp_json_encode` does sanity checking on the UTF-8 encoding of strings, so probably good to keep that?

joehoyle:
@rmccue I think using it is probably still ok, but I'm not sure we should be disallowing the internal PHP functions.

roborourke:
There are now cases where our coding standards should apply that are either not in the WP context or too early in the WP bootstrapping process for this to be possible.

kadamwhite:
@joehoyle @roborourke If one of you can fix the merge conflict, we'll get this landed

kadamwhite:
Went ahead and cleaned that up myself; some of the other PR's were confusing Git's normal merge algorithm something fierce. Will land once checks pass.

Issue #89, opened 2018-09-07
Support [namespace].php too in file naming convention


joehoyle:
I think I might have seen this proposal somewhere before, but I can't find it here. Instead of requiring `my_namespace/namespace.php` where your namespace doesn't have any classes, we should support `my_namespace.php` so there's a little less folder cruft.

mikeselander:
Related is: https://github.com/humanmade/Engineering/issues/83

rmccue:
Merged in https://github.com/humanmade/Engineering/pull/112 so we need to update our rule here.

PR #99, opened 2018-10-15
Allow $namespace.php in function files

joehoyle:
See https://github.com/humanmade/coding-standards/issues/89

Needs some tests too, putting here as a first pass.


mikeselander:
FYI https://github.com/humanmade/coding-standards/blob/master/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php#L86 will also need to be updated as it will push for a `/directory/namespace.php` in the logic and complain if you're in a subdirectory off of `/inc` and have a non-standards file name.

For example:
`/inc/my-custom-namespace` -> passes due to a short circuit in the code
`/inc/my-custom-namespace.php` -> passes
`/inc/bananas/my-custom-namespace.php` -> fails

joehoyle:
👍

mikeselander:
@joehoyle do you think you'll have time to finish this up over the next few weeks or should someone else pick this up?

joehoyle:
I'm pretty busy with the Altis version 1 timeline, if anyone else has time to pick this up within the next month, please go ahead :)

mikeselander:
@rmccue this is ready for a review, would appreciate you eyes on this 🙏

PR #103, opened 2018-11-14
Stylelint: Define `function-url-quotes`

mikeselander:
Stylelint is currently (via a parent config, I assume) requiring that _no_ quotes be used within the CSS `url()` function. However, using `url()` without quotes requires special escaping and the potential for forgetting to escape a special character exists, making use with quotes just a little bit safer. I propose that we require quotes in URLs for consistency and ease of use.

Ref: https://drafts.csswg.org/css-values-3/#urls

This PR defines this rule to "always" to match the current spec.

### DIFF
--- a/packages/stylelint-config-humanmade/stylelint.json
+++ b/packages/stylelint-config-humanmade/stylelint.json
@@ -18,6 +18,7 @@
     },
     "function-parentheses-space-inside": "always-single-line",
     "function-comma-space-after": "always-single-line",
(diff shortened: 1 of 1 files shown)


rmccue:
Should this be added to the [style guide](https://github.com/humanmade/Engineering/blob/master/standards/style/css/index.md) as well?

mikeselander:
Yes, it and #104 should both be updated in the guides. I'll add PRs for those as well.

Issue #104, opened 2018-11-14
Stylelint: Consider defining Specific Properties Available


mikeselander:
We are currently whitelisting unit(less, actually)s for `line-height` only in our stylelint config, but I propose that we get more specific with what units are available.

### Font Size
Common best practice is to use `rem` units for font size based off of a single root value. We should enforce `rem` units, or at a minimum, blacklist `px` units to prevent them from being used.

### `ms` Units
For timing functions, such as transitions or animations, `ms` units are far clearer and easy to digest than seconds because they are laid out in longform. While they mean the same thing, `400ms` is cleaner than `.4s` and feels more precise in my opinion. I suggest that we enforce ms units on all timing functions.

Ref https://stylelint.io/user-guide/rules/declaration-property-unit-whitelist/ & https://stylelint.io/user-guide/rules/declaration-property-unit-blacklist/

Thoughts #teamfrontend?

cc @peterwilsoncc @sambulance @goldenapples @rmccue @joemcgill

peterwilsoncc:
For font-size: I'd prefer to blacklist `px` rather than whitelist `rem`. `em` can be [useful within components](https://css-tricks.com/rem-global-em-local/) and other relative units such as `vw, vh` can be useful for responsive font sizes.

Timing functions: `ms` works for me.

I'm going to cc @kirstyburgoine as part of #teamfrontend

rmccue:
I strongly agree that we should block `px` rather than only allow `rem`. The bane of my existence right now is the overuse of `rem` within the pattern library, which makes it much harder to use things like headers inside sidebars where the whole sidebar is `0.9rem`.

IMO, `rem` should be used for layout-level things, like sizing up/down sidebars, content areas, etc as well as page layout. `em` should be used for elements within blocks. A different way to think about this could be that `rem` should only be used with ID selectors, and `em` should be used elsewhere. (Although, I'd actually use classes rather than IDs typically...)

Agree on `ms`, and I haven't actually seen `s` much in the wild.

p.s. created a @humanmade/frontend team

kirstyburgoine:
Thanks for cc'ing me @peterwilsoncc :) I definitely agree with blacklisting `px` as the approach. 

I think my approach to units is a bit different to @rmccue's. I use `rem` for all type sizing. This is because it takes its base value from the body font size making it an easier unit to work with. It is also needed when applying responsive type calculations to the body tag so that they are applied to all type sizing in your page. 

[...]

PR #106, opened 2018-11-16
Unit Listing Requirements

mikeselander:
As discussed in #104, requiring that `ms` units be used for timing durations and ensure that `px` values cannot be used for font-sizes.

Have I missed any timing durations in this list?

cc @humanmade/frontend for thoughts/review.


mikeselander:
Good catch @sambulance - added those in 👍

PR #112, opened 2018-12-17
Add `NamespaceDirectoryName Exclusion` for `/tests` directories

### LINKED ISSUE #92: Ignore `inc` namespace rule in test directories
The `NamespaceDirectoryNameSniff` sniff currently requires namespaces within test directory files, but this should not be required as test files are outside of a plugin-based organizational structure.

Should check for `/.test` and `/test` directories in path and if found, bypass this rule.

mikeselander:
Matching #92, ignoring the inc directory rule when parsing `/tests` directories.

Resolves #92


mikeselander:
@joehoyle I think it's possible but we don't exclude any of our other internal rules with that method and we're already excluding another directory in this rule's code. It would make it very difficult to tell where exclusions existed if we started breaking that up now.

joehoyle:
Ok makes sense

PR #113, opened 2018-12-18
Add Codesniffer Installer

### LINKED ISSUE #100: Switch to DealerDirect/phpcodesniffer-composer-installer
https://github.com/DealerDirect/phpcodesniffer-composer-installer is a better way of setting `installed_paths` than the hack we have. Should fix #2 in the process.

mikeselander:
Switching to using the ` DealerDirect/phpcodesniffer-composer-installer` package for finding configuration paths.

I did not use any [direct calls](https://github.com/DealerDirect/phpcodesniffer-composer-installer#calling-the-plugin-directly) as this appears to work properly as-is.

Resolves #100


roborourke:
@mikeselander can you resolve the conflicts here?

PR #116, opened 2018-12-18
Set Composer Library Type

mikeselander:
For the purposes of Composer, these coding standards are a PHPCS standard and currently we've only marked this as a "project". We should be more specific about the type of library this is for sorting and path-directing purposes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
 	"name": "humanmade/coding-standards",
 	"description": "Human Made coding standards",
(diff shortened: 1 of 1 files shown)


mikeselander:
> The plugin that provides this type need be installed too?

To take advantage of this specific type, a developer would have to install the correct dependency to handle it alongside this. Without the extra installer types handling, Composer just puts it in `vendor` alongside everything else.

PR #121, opened 2019-02-14
Require space in curly braces for children

mattheu:
Currently this code is allowed. I think we should enforce a space inside curly braces for children.

```
<div>
	{children}
	{foo && (
		<p>Foo</p>
	)}
</div>
```

Expected:

```
<div>
	{ children }
	{ foo && (
		<p>Foo</p>
	) }
</div>
```


kadamwhite:
Landing this after discussion in Slack indicates there's no reason to continue holding off :smile:

PR #125, opened 2019-03-22
Add Changelog file

mikeselander:
Adds a changelog file so that we can update it as we go and so that there is some reference for changes. This also moves the current changelog from my notes to a place that anyone can leverage it :) 

I've attempted to backfill this from old releases but it might have to be adjusted.

I think that it might be a good idea to enforce that this file be updated for every pull request on future releases. This would allow the changelog to fill itself in and releases would be less work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,13 +50,37 @@
 
 ## 0.5.0 (May 22, 2018)
 
+- Update ESLint config peer dependencies #65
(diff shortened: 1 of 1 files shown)


mikeselander:
> I don't know how we'd go about doing this but I would be extremely in favor!

I think that we could add a PR template with a checklist listing this file as one of the items.

PR #126, opened 2019-03-22
Remove Composer References in NPM Packages

mikeselander:
We've had a line recommending usage via Composer in both NPM packages but this it no longer recommended. Removing for future users.


kadamwhite:
Merging as-is despite failures, since the change is docs only

PR #127, opened 2019-03-23
Restore `allowMultipleArguments`

mikeselander:
In WPCS 1.1.0, the `allowMultipleArguments` property from `PEAR.Functions.FunctionCallSignature` was disallowed, causing our tests to fail and existing behavior to be altered. 

I re-allowed this to restore expected behavior and updated Composer to more explicitly show which version we are on (`^1.0.0` was already giving us 1.2.1, just without the explicit declaration).

We can easily turn this back off if we decide to alter the behavior in the future. I believe this to be the only "breaking"/expectation-changing behavior from WPCS 1.0.0 -> 1.2.0.

### DIFF
--- a/HM/ruleset.xml
+++ b/HM/ruleset.xml
@@ -89,6 +89,25 @@
 
 	<rule ref="WordPress.WP.TimezoneChange" />
 
+    <!--
+    Restore the ability to have multiple arguments per line
+
+    WPCS disallowed this behavior in 1.1.0, but we'd like to keep it until
+    there is a reason to disallow this behavior.
+    Ref: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/commit/bb8a48671e213a5588a6439ea52411eeefab4b0f
+    -->
+    <rule ref="PEAR.Functions.FunctionCallSignature">
+        <properties>
+            <property name="allowMultipleArguments" value="true" />
+        </properties>
+    </rule>
+    <rule ref="PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket">
+        <severity phpcs-only="true">0</severity>
+    </rule>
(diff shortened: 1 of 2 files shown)


mikeselander:
Agreed, I don't have a problem with their philosophy. I'm more afraid of this throwing a bunch of new flags all of a sudden for developers without greater discussion of the topic. I'll open an Issues for this.

composer.json
```diff
@@ -5,7 +5,7 @@
 	"license": "GPL-2.0-or-later",
 	"require": {
 		"php": ">=7.1",
-		"wp-coding-standards/wpcs": "^1.0.0",
+		"wp-coding-standards/wpcs": "^1.2.0",
```
joehoyle: Perhaps we should use `~` here instead? I _think_ `wpcs` makes changes in minor versions, that we might be better off pulling in explicitly? I'm mostly in favor of more restrictive constraints thought.

mikeselander: If we're going to be more restrictive, maybe we should do `1.2.*` instead 🤔 

I'm also in favor of more restrictive versions.

joehoyle: Isn't that what `~1.2.0` means?

rmccue: I would actually peg it specifically to 1.2.whatever, because wpcs also makes changes in bugfix releases occasionally.

HM/ruleset.xml
```diff
@@ -89,6 +89,25 @@
 
 	<rule ref="WordPress.WP.TimezoneChange" />
 
+    <!--
```
rmccue: This is using spaces for indentation rather than tabs.

mikeselander: 🤦‍♂️

mikeselander:
@rmccue I've adjusted the spaces - any more thoughts on this change?

PR #137, opened 2019-05-07
Change Stylelint Directory

mikeselander:
Changes the Stylelint directory location to match the correct directory in the final NPM location `@humanmade/stylelint-config`. 

This breaks with the `eslint-config-humanmade` naming from eslint, but allows us to place the package underneath the Human Made organization. This gives more administrators proper access and improves discoverability of the package.


packages/stylelint-config/readme.md
```diff
@@ -9,14 +9,14 @@ This package is a Stylelint shareable configuration, and requires the `stylelint
...
-npm install --save-dev stylelint stylelint-config-humanmade
+npm install --save-dev stylelint @humanmade/stylelint-config
 ```
 
 Then, add a `.stylelintrc` file and extend these rules. You can also add your own rules and overrides for further customization.
 
 ```json
 {
-  "extends": "stylelint-config-humanmade",
+  "extends": "@humanmade/stylelint-config/stylelint.json",
```
ntwb:
```suggestion
  "extends": "@humanmade/stylelint-config",
```

The `styleint.json` is not required

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Changed
  - Exclude `load.php` from `NamespaceDirectoryNameSniff` #131
  - Allow `json_encode` / `json_decode` function usage #97
+ - Updated location of Stylelint package to reflect correct NPM name #137
```
ntwb:
```suggestion
 - Updated location of stylelint package to reflect correct NPM name #137
```

A nit, stylelint is spelled with all lowercase 😏

packages/stylelint-config/readme.md
mikeselander: Thank you for the review @ntwb 🙏 I tested this out on a client project (I published this privately for testing) and found that stylelint couldn't find the config without specifying the actual file path. 

I kept getting `Error: Could not find "@humanmade/stylelint-config".` when using the following:

```json
{
	"extends": "@humanmade/stylelint-config"
}
```

However, the verbose option with the filename worked.

I assumed this was because we have no entry `index.js` files on the project. Any thoughts on what I was doing wrong?

ntwb: right, thanks, stylelint uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to handle traversing the directory tree in search of the stylelint config file to be used, as such rather than having the file named `stylelint.json` it should be `.stylelintrc.json`, or any of the other supported _rc_ filename types supported by cosmiconfig as documented here: https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#loading-the-configuration-object

packages/stylelint-config/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@humanmade/stylelint-config",
+  "description": "Human Made coding standards for CSS and SCSS.",
+  "version": "0.6.0",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/humanmade/coding-standards.git"
+  },
+  "devDependencies": {
+    "stylelint-config-wordpress": "^13.0.0"
```
ntwb:
```suggestion
    "stylelint-config-wordpress": "^14.0.0"
```

p.s. v14 was released a couple of weeks ago https://github.com/WordPress-Coding-Standards/stylelint-config-wordpress/releases/tag/14.0.0

rmccue:
The ESLint package is called that so it works with `eslint -c humanmade` I think; do similar restrictions apply with Stylelint?

CHANGELOG.md
mikeselander: Got it , thank you ❤️

ntwb:
> The ESLint package is called that so it works with `eslint -c humanmade` I think; do similar restrictions apply with Stylelint?

No, this cannot be done with stylelint at this time:

See https://github.com/stylelint/stylelint/issues/3876

rmccue:
Brilliant, in that case, `@humanmade/stylelint-config` makes a lot of sense 👍

packages/stylelint-config/package.json
mikeselander: Good note, I'll handle that in a separate PR 👌

packages/stylelint-config/readme.md
mikeselander: Updated to `.stylelintrc.json`

PR #139, opened 2019-05-13
Test Suite Improvements

mikeselander:
While debugging #81, I made a series of style fixes that keep the code a bit cleaner and easier to read.

 - Removes unused `use` statements
 - Moves classes from `\` to `use`
 - Defines class variables used by the class
 - Adds `@group` definitions for more finite testing
 - Defines some constants that should help PHPCS work more intelligently while testing

### DIFF
--- a/tests/AllSniffs.php
+++ b/tests/AllSniffs.php
@@ -11,7 +11,6 @@ namespace HM\CodingStandards\Tests;
 
 use PHP_CodeSniffer\Util\Standards;
 use PHP_CodeSniffer\Autoload;
-use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
 use PHPUnit\TextUI\TestRunner;
 use PHPUnit\Framework\TestSuite;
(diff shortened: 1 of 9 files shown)


tests/AllSniffs.php
```diff
@@ -44,8 +46,8 @@ public static function suite() {
 		$suite = new TestSuite( 'HM Standards' );
 
 		$standards_dir = dirname( __DIR__ ) . '/HM';
-		$all_details = Standards::getInstalledStandardDetails( false, $standards_dir );
-		$details = $all_details['HM'];
+		$all_details   = Standards::getInstalledStandardDetails( false, $standards_dir );
+		$details       = $all_details['HM'];
```
rmccue: Didn't we remove the equals-align rule?

mikeselander: Indeed we have, I hadn't setup PHPStorm properly so it was using the defacto set. I can change these back, I suppose.

PR #140, opened 2019-05-14
Fix NamespaceDirectoryNameSniff for tests directory length

### LINKED ISSUE #134: tests/ not correctly exempt from HM.Files.NamespaceDirectoryName
The [logic in HM.Files.NamespaceDirectoryName](https://github.com/humanmade/coding-standards/blob/master/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php#L67-L68) to exclude the 'tests' directory from the sniff has a logic bug, causing the sniff to fail with the error
```
Directory ts for namespace RootNS\Tests found; use tests instead
```
The offending lines are the `substr` command which uses `strlen( '/inc' )` instead of detecting whether it should be trimming `/inc` or `/tests`:
```
	$inc_position = $matches[0][1] ?? $test_matches[0][1];
	$after_inc = substr( $directory, $inc_position + strlen( '/inc' ) );
```
A regex or more conditional `substr` call should resolve the issue.

mikeselander:
Fixes #134

 - Adds tests for the `tests` directory that prove the initial failure
 - Fixes the code parsing for the incorrect directory length that triggered the `Directory ts for namespace` error


HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
```diff
@@ -64,15 +64,18 @@ public function process( File $phpcsFile, $stackPtr ) {
 			return;
 		}
 
-		$inc_position = $matches[0][1] ?? $test_matches[0][1];
-		$after_inc = substr( $directory, $inc_position + strlen( '/inc' ) );
-		if ( empty( $after_inc ) ) {
+		// Find correct after namespace-base path.
+		$after_dir = ! empty( $matches[0][1] )
+			? substr( $directory, $matches[0][1] + strlen( '/inc' ) )
+			: substr( $directory, $test_matches[0][1] + strlen( '/tests' ) );
```
rmccue: Not a big fan of this style. Can we instead set the offset on a separate line which can fit on one line, and `substr` once?

mikeselander: We can, but I think the alternative is quite a bit messier.

Something like:

```php
// Find correct after namespace-base path.
$match = $matches[0][1] ?? $test_matches[0][1];
$offset = $matches[0][1] ? strlen( '/inc' ) : strlen( '/tests' );
$after_dir = substr( $directory, $match + $offset );
```

Unless I'm missing a better way of doing this, we end up with two different if blocks for the same logic, and combining them into a proper `if/else` statement would mean setting the same temporary variable twice.

Thoughts?

rmccue: Yeah, I would do it that way, as it shortens the lines substantially.

Alternatively, change the regex to `'#/(inc|tests)(?:/|$)#'` and use `substr( $directory, $matches[0][1] + $matches[1][1] )`

tfrommen: Not sure I get this right, but I find both the original and the suggested code a bit overly complex.

Isn't all we need the path after `/inc` or `/tests`? And fail if there is no such path (meaning the code does not live in such a folder)?

How about changing the code like so:

```php
		if ( ! preg_match( '|(?:/inc|/tests)(/.*)?|', $directory, $matches ) ) {
			$error = 'Namespaced classes and functions should live inside an inc directory.';
			$phpcsFile->addError( $error, $stackPtr, 'NoIncDirectory' );
			return;
		}

		$after_dir = $matches[1] ?? '';
```

Done.   ...? 🤔 😀

mikeselander: @tfrommen you're not wrong - they are overly complex 💯. I was planning on using a single regex like Ryan suggested but hadn't gotten to fixing it like you have here yet.

That stored sandbox link is really neat, I've never seen that before!

mikeselander:
I simplified the logic above and modified the regex to look for the last available instance of `inc` or `tests` in cases (like our test files) where there are two. I expect our tests are the only case this happens but it felt worth covering the case.

HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
```diff
@@ -55,24 +55,22 @@ public function process( File $phpcsFile, $stackPtr ) {
 			return;
 		}
 
-		if (
-			! preg_match( '#/inc(?:/|$)#', $directory, $matches, PREG_OFFSET_CAPTURE )
-			&& ! preg_match( '#/tests(?:/|$)#', $directory, $test_matches, PREG_OFFSET_CAPTURE )
-		) {
+		if ( ! preg_match( '#(?:.*)(?:/inc|/tests)(/.*)?#', $directory, $matches, PREG_OFFSET_CAPTURE ) ) {
```
tfrommen: Why do you still use `PREG_OFFSET_CAPTURE`? You don't need/use the index any longer. This would also simplify the new line 65 from `$after_dir = $matches[1][0] ?? '';` to `$after_dir = $matches[1] ?? '';`.

Or is there something else going on that I don't see?

PR #149, opened 2019-07-25
Add typescript documentation

nathanielks:
Nothing much else to say! Out of the box, the eslint package will mostly work fine, but it will throw errors some TypeScript-specific syntax. This fixes that!

### DIFF
--- a/packages/eslint-config-humanmade/readme.md
+++ b/packages/eslint-config-humanmade/readme.md
@@ -26,6 +26,22 @@ extends:
 - humanmade
 ```
 
+### Working with TypeScript
+
+If you desire to use TypeScript for your project, you will need to add another dependency:
+
+```
+npm install --save-dev @typescript-eslint/parser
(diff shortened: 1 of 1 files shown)


nathanielks:
Aw, damn:
```
Failed to run hmlinter
Could not run: Error: Cannot find module '@typescript-eslint/parser'
```

Is there a way we can have hmlinter install linter dependencies?

pdewouters:
@nathanielks I believe you can make a PR to add them here to our `eslint-config-humanmade` packages

kadamwhite:
@nathanielks The docs look good; lookin at your "Oh, Damn", do you mean that it still doesn't work if you install that package, as you documented?

nathanielks:
@kadamwhite it's been so long I don't recall 🙈 I did some local testing and I cannot reproduce the issue, but if I do I'll follow up with another PR 👍

PR #150, opened 2019-07-25
Add npx install-peerdeps documentation

nathanielks:
This adds a simple shortcut for installing peerdeps.


nathanielks:
Updated! I left the AirBNB credit because that's also where I found out about `npx install-peerdeps`!

PR #155, opened 2019-12-02
Make react hooks and sonarjs eslint plugins available

pdewouters:
## Description

Given I want to use some extra linting plugins in my config, and the project is using the HM eslint config, the way to use these extra plugins, and have HM linter bot use them in its review, I need to include these plugins in this `package.json` file.

I'd like to be able to lint my React code with the plugin that enforces the rules of hooks

https://reactjs.org/docs/hooks-rules.html

I want to use the linting rules provided by this package
https://github.com/SonarSource/SonarJS  

This is the accepted way to add custom plugins as was discussed with @rmccue

PR #156, opened 2019-12-04
Add typescript dev dependency

pdewouters:
## Description

Adds the typescript npm package
https://www.npmjs.com/package/typescript

## Motivation and Context

Having the ability to use typescript types checking will allow projects to benefit from stricter type checks in Javascript code without having to write Typescript.

This is supported by the fact that the Gutenberg team are doing the same: https://github.com/WordPress/gutenberg/commit/f7e067b2a08b75a7dd889f87473f3e434153c743#diff-e5e546dd2eb0351f813d63d1b39dbc48

This can be paired with the `eslint` rules `valid-jsdoc` and `require-jsdoc`

To use in a project, developers will simply need to add a `tsconfig.json` file, which should have the key `"noEmit": true,` and run `tsc` on their codebase

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/packages/eslint-config-humanmade/package.json
+++ b/packages/eslint-config-humanmade/package.json
@@ -16,7 +16,8 @@
     "eslint-plugin-flowtype": "^3.2.0",
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jsx-a11y": "^6.1.2",
-    "eslint-plugin-react": "^7.11.1"
+    "eslint-plugin-react": "^7.11.1",
+    "typescript": "^3.7.3"
   },
   "peerDependencies": {
     "babel-eslint": "^10.0.0",


ntwb:
Yes please, similar to #155 could we add some basic docs in the readme/s that the package includes this package a link to the docs of the package please

kadamwhite:
After discussion with @rmccue and @mikeselander we don't think it makes sense to include the typescript compiler inside the coding standards repo. We can flesh the documentation out to clarify what dependencies a project needs to declare in order to utilize TS-specific rules, but having the compiler bundled with the standards feels like it's conflating two different things.

PR #151, opened 2019-08-06
Update: WPCS 2.1.1 / VIPCS 2.0.0

pdewouters:
Updates WP CS and VIP CS packages to latest versions.
VIP CS rules are now compatible with WP CS 2.*
see https://github.com/Automattic/VIP-Coding-Standards/releases/tag/2.0.0

Resolves #130 
Resolves #148

PR #163, opened 2019-12-27
Limit Fixture Test Standards

mikeselander:
**Problem:**
We've been loading any and all standards from our HM ruleset within our fixture tests, but the fixture tests only store errors for our specific standards. Loading in other standards is causing all kinds of false errors such as those seen in #81 and #151. I believe that these errors are related to PHPCS autoloading some classes and then our standard coming in and triggering them to load again. PHPCS runs fine and as expected on the command line, but fails in the test suite for the above hard-to-trace reasons. 

**Solution:**
The Ruleset class offers us the ability to limit the standards loaded to a specific subset so that our tests run only against our custom code. Since we're only recording errors on our custom sniffs, there's little point to loading in all the standards for these particular tests. It also cuts the run time for the test suites down by half. Yay 200ms back in your life ;P  

Note that this does remove the ability to smoke test for regressions from updating external rulesets, but I think that we should handle that in a separate test suite anyway.

This fixes the test suite issues seen in both #81 and #151 as seen by these test branch CI runs: https://travis-ci.org/humanmade/coding-standards/builds/630048500 and https://travis-ci.org/humanmade/coding-standards/builds/630048408

https://github.com/squizlabs/PHP_CodeSniffer/blob/278b643f2a5ec1b92fd60b2d7dac034f3a52ea4e/src/Ruleset.php#L177-L191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Updated
  - Bumped `stylelint-config-wordpress` package to v15 from v13 #165
  - Ignore stylelint `at-rule` line break for `if/else/elseif` #170
+ - Restricted fixture tests to load only custom HM sniffs #163
 
 ## 0.7.0 (June 5, 2019)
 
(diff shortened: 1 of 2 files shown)


mikeselander:
> Maybe we should consider marking specific sniffs in each fixture file?

This makes sense to me; I'll add an issue for follow-up on that.

PR #164, opened 2019-12-27
Disallow Usage of `!important` in CSS

### LINKED ISSUE #144: Consider Flagging Usage of `!important` in Stylelint
While `!important` _can_ be used intelligently and useful in specific situations, it's generally bad practice. I know that we don't want to specifically block all usages of `!important`, but I think it would be beneficial to make a developer justify each usage of it.

I recommend that we adapt the [existing `!important` rule](https://github.com/stylelint/stylelint/blob/master/lib/rules/declaration-no-important/index.js) (or just use it as-is) to display a message requiring a justifying comment to dismiss the error. In this way, a reviewer can review the logic for using `!important` which should allow them to offer alternatives and hopefully reduce the usage of `!important`.

Thoughts @humanmade/frontend?

mikeselander:
This PR adds the stylelint [`declaration-no-important`](https://stylelint.io/user-guide/rules/declaration-no-important) rule which disallows usages of `!important` unless a developer specifically adds a exception comment.

Fixes #144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased (0.8.0)
 
+### Added
+ - Disallowed usage of `!important` in CSS
+
 ## 0.7.0 (June 5, 2019)
 
 ### Changed:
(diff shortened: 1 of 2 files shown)


mikeselander:
As a note, we should update the Handbook in tandem with this change.

PR #168, opened 2019-12-28
Add Rule to Detect Consecutive Newlines

### LINKED ISSUE #60: Detect consecutive blank lines
Similar to https://github.com/wikimedia/mediawiki-tools-codesniffer/commit/3a6709be2612fad63b9e9ead4e6644c28748edcc

mikeselander:
This PR copies in a rule to handles multiple empty lines all throughout PHP. I copied this rule in the MediaWiki ruleset as it felt unnecessary to include their whole ruleset for one very small rule and this is a particularly maintainable sniff. I am open to opinions on the approach.

This will detect any empty consecutive lines. It improves on the main Squiz rule as the Squiz rule only detects multiple lines within functions, and not in classes or in general PHP.

Fixes #60

### DIFF
--- a/HM/Sniffs/Whitespace/MultipleEmptyLinesSniff.php
+++ b/HM/Sniffs/Whitespace/MultipleEmptyLinesSniff.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * Check multiple consecutive newlines in a file.
+ */
+
+namespace HM\Sniffs\Whitespace;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+/**
+ * Class that checks for more than one consecutive empty line.
+ *
+ * This sniff is adapted from the MediaWiki Tools ruleset
+ * See: https://github.com/wikimedia/mediawiki-tools-codesniffer/commit/3a6709be2612fad63b9e9ead4e6644c28748edcc#diff-4522e134348f6d4e2efe9ccb7148a254
+ */
+class MultipleEmptyLinesSniff implements Sniff {
+
+	public function register() {
+		return [ T_WHITESPACE ];
+	}
+
+	public function process( File $phpcsFile, $stackPtr ) {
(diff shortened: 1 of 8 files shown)


HM/Sniffs/Whitespace/MultipleEmptyLinesSniff.php
```diff
@@ -0,0 +1,60 @@
...
+	public function register() {
+		return [ T_WHITESPACE ];
+	}
+
+	public function process( File $phpcsFile, $stackPtr ) {
+		$tokens = $phpcsFile->getTokens();
+
+		if ( $stackPtr > 2
+			&& $tokens[$stackPtr - 1]['line'] < $tokens[$stackPtr]['line']
+			&& $tokens[$stackPtr - 2]['line'] === $tokens[$stackPtr - 1]['line']
```
rmccue: Needs spacing inside `[]`; assuming this is the original formatting from MW?

Also, these conditionals should be flipped and broken up.

mikeselander: Ah ya, they were pulled over directly from the old rule. I'll clean them up a bit

mikeselander: What, exactly, would you like flipped and/or broken up in these conditionals @rmccue?

rmccue:
```
if ( $stackPtr > 2 && $tokens ... ) {
    do_the_thing();
```
➡️ 
```
if ( $stackPtr <= 2 ) {
    return;
}

if ( $tokens ... ) {
    return;
}
```

This should also reduce the line length significantly and mean you don't need to wrap them. (Of course, you can still combine the conditions, but often better to break them up for readability if they're this long.)

mikeselander: Gotcha, so early returning, that makes more sense. Thanks for clarifying!

PR #169, opened 2020-01-02
ESLint Enforce semicolons

### LINKED ISSUE #153: JS: Enforce trailing semicolons?
Our linter doesn't pick up on trailing semicolons. This can lead to inconsistency inside JS files:

```javascript
const showSection = section => {
	section.setAttribute( 'aria-expanded', true )
}

const hideSection = section => {
	section.setAttribute( 'aria-expanded', false );
};
```

Both of the above are valid. We should enforce either one way or the other.

mattheu:
Enforce semicolons.

Closes #153

### DIFF
--- a/packages/eslint-config-humanmade/fixtures/fail/semicolon.js
+++ b/packages/eslint-config-humanmade/fixtures/fail/semicolon.js
@@ -1,0 +1,14 @@
+/* eslint-disable no-unused-vars */
+
+// Missing semicolon.
+const testMissingSemicolon = 'foo'
+
+// Missing semicolon.
+const testSemiSpacing = 'bar' ;
+
+// Double semicolon.
+const testDoubleSemiColon = 5;;
+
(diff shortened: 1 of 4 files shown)


mikeselander:
As a note, we should update the Handbook in tandem with this change.

packages/eslint-config-humanmade/fixtures/fail/semicolon.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-unused-vars */
+
+// Missing semicolon.
+const testMissingSemicolon = 'foo'
+
+// Missing semicolon.
```
sambulance: The semicolon isn't missing here, just spaced incorrectly.

mattheu: Oh thanks Sam - I'll update the comment

Issue #85, opened 2018-08-22
Split standard into core and style


rmccue:
In order to better assess third-party code, we should split out the core (security, performance) rules from the style rules. We should do this for both PHP and JS, but PHP is a larger priority.

kadamwhite:
This is complete for PHP following #180 ! :tada: :confetti_ball: 

JS and CSS would both benefit from this treatment, but after discussion with @rmccue that's a lower priority and we can open new issues to track those backlog items in the future. Closing this out for 1.0.

Issue #130, opened 2019-04-02
Update WPCS to 2.1.0


mikeselander:
We recently bumped WPCS to 1.2.1, but it took so long that they've now released a version 2.1.0. We should probably look at bumping that version up for our standards.

ntwb:
FYI: WPCS 2.1 is planned for release April 8th

mikeselander:
Thanks for the FYi @ntwb, updated 👍

mikeselander:
Noting that VIPCS conflicts with WPCS 2.*, which would affect our ability to upgrade. It looks like the VIPCS team is working on compatibility but there's no promise when it would be ready: https://github.com/Automattic/VIP-Coding-Standards/milestone/4

spacedmonkey:
At the same time as looking into this, you will likely need VIP-Coding-Standards to 1.0.0 . See #136

spacedmonkey:
Any plans for this?

mikeselander:
@spacedmonkey we will definitely upgrade to this at some point, but we're not in a rush here. These upgrades tend to break existing expectations on current projects so we take them pretty slowly. Additionally there is a dependency on VIPCS updating their code to work with 2.* of WPCS so we won't even consider this until that's fixed.

That being said, if someone else has the time to look into and consider regressions that might come from the 1.2 -> 2.1 upgrade we're always accepting PRs!

spacedmonkey:
Now that VIP GO coding standard is compatible with VIP GO standard 2.0. These is no reason not to update them both.

rmccue:
This will include https://github.com/WordPress/WordPress-Coding-Standards/pull/1684 too.

ntwb:
I've updated #151 and it now passes with the latest release of WPCS & VIPCS

PR #180, opened 2020-02-12
Add a HM Minimum standard

rmccue:
Opening a PR to track the work in progress on splitting security/performance required checks from our style checks. See #85.

The aim is for all code deployed to Altis Cloud to pass these checks, so it should contain no style-based issues. We'll also eventually check dependencies against this.

Work ongoing here.


rmccue:
To compare the rules, run the following:

```
vendor/bin/phpcs -e --standard=HM
vendor/bin/phpcs -e --standard=HM-Required
```

This will print which sniffs are included in each ruleset. (You might want to run the first one on `master` to compare, as the HM ruleset extends HM-Required here.) I used this to generate [a diff](https://gist.github.com/rmccue/541e378642f4c7a44c0ed1f7c2dc9d02); the first file there is `HM-Required` on this branch vs `HM` on master, the second is `HM` on this branch vs `HM` on master.

(cc @mikeselander)

rmccue:
Also, we're going to have to subclass `EscapeOutputSniff.php` to override the `trigger_error`/`error_log` warnings. I filed an issue in WPCS to add these as config options or include the function in the error code, but seems there's little appetite for that: https://github.com/WordPress/WordPress-Coding-Standards/issues/1864

rmccue:
> Also, we're going to have to subclass `EscapeOutputSniff.php`

This will have a back-compat break btw, because the name of the sniff in ignores will need to change from `WordPress.Security.EscapeOutput.OutputNotEscaped` to `HM.Security.EscapeOutput.OutputNotEscaped`. 😞 As far as I can tell, phpcs has no mechanism to alias this.

rmccue:
Remaining errors I think we need to fix (I'll add to this list as I go):
* [ ] Querying by `meta_key` with `EXISTS` is incorrectly marked as slow
* [ ] Usage of null-coelescing operator isn't handled correctly (may require #130)
* ...

rmccue:
@joehoyle Can I get an initial review of the rules in here?

joehoyle:
I wonder if we should use the Altis namespace rather than HM here. I think it follows well that "required" things are in Altis, and style / best practices are in HM.

rmccue:
Yeah, it's called this for convenience right now, but we can look at renaming later.

rmccue:
OK, I think this is now ready for a proper review @joehoyle @mikeselander. It's probably worth pulling this down and running against a project to test; pull the branch, `composer install`, then `vendor/bin/phpcs --standard=HM-Required ../project/path`

We may want to change some of these or override them for the main HM standard, as there's a few exclusions which are looser than we might want to recommend. I'll see about going through those, but the main standard should be OK now.

I think we want to extend this with an "Altis" standard that includes this but overrides a bunch of the messages to point to Altis-specific documentation. We might want to include that here too, but not sure.

joehoyle:
Few things I'm seeing, which might want addressing:

- twentyseventeen is failing a lot of checks, might be worth checking the default themes too. I feel like we probably shouldn't be failing those as I'd _expect_ it's not the theme that's wrong, its our static analysis.

- I'm seeing a bunch of ` 600 | ERROR | 'object' type declaration is not present in PHP version 7.1 or earlier`, did I misconfigure something? I wouldn't expect to see this.

- Getting a few ` 189 | ERROR   | Missing wp_unslash() before sanitization.`, is that actually a security issue? Or is this just a bug/unexpected functionality sniff?

- On slow meta queries, I had `'meta_key'     => 'deploy_id', // WPCS: slow query ok.` does this style of exclude no longer work?

- I'm getting "All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks) | found 'str_repeat'" for `echo ':' . str_repeat( ' ', 2048 ) . "\n\n";` static analysis should be able to see there's no dynamic data here?

- It's showing errors for the unit test files in `./tests` maybe we should add a default exclude for that?

In total is shows 82 errors and 6 warning across 25 files.

rmccue:
By the way, this is now set up with the altis-review bot at https://github.com/humanmade/altis-review-test if you want to send test PRs there.

HM/Sniffs/Performance/SlowMetaQuerySniff.php
```diff
@@ -0,0 +1,315 @@
...
+	protected $stackPtr;
+
+	/**
+	 * Groups of variables to restrict.
+	 *
+	 * @return array
+	 */
+	public function getGroups() {
+		return array(
+			'slow_query' => array(
```
mikeselander: This is a general sniff-writing question: what is the reason we tend to write these with long-form array keys? We don't write any other code assuming that it won't be PHP 5.4+. I've seen this on most other sniffs as well.

HM/Sniffs/Performance/SlowMetaQuerySniff.php
```diff
@@ -0,0 +1,315 @@
...
+namespace HM\Sniffs\Performance;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Util\Tokens;
+use WordPress\AbstractArrayAssignmentRestrictionsSniff;
+
+/**
+ * Flag slow meta queries.
+ */
+class SlowMetaQuerySniff extends AbstractArrayAssignmentRestrictionsSniff {
```
mikeselander: Is this all custom code you've written or  plucked and modified from somewhere? If plucked, it would be useful to have a note on where from and why/what was modified.

HM-Required/ruleset.xml
```diff
@@ -0,0 +1,307 @@
...
+	<!-- Remove after https://github.com/squizlabs/PHP_CodeSniffer/pull/1664 -->
+	<rule ref="WordPress.PHP.DiscourageGoto"/>
+	<rule ref="WordPress.PHP.DiscourageGoto.Found">
+		<type>error</type>
+		<message>The "goto" language construct should not be used.</message>
+	</rule>
+
+	<!-- Disallow cron checks which are too frequent. -->
+	<rule ref="WordPress.WP.CronInterval" />
+	<rule ref="WordPress.WP.CronInterval.CronSchedulesInterval">
```
mikeselander: This prevents cron intervals less than 15 minutes, is that something we really want? I've seen a lot of good reasons to have more frequent intervals and 15 minutes is quite long. Perhaps this could be a minimum of 5 minutes instead?

HM-Required/ruleset.xml
```diff
@@ -0,0 +1,307 @@
+<?xml version="1.0"?>
+<ruleset name="HM-Required">
```
mikeselander: I'm not keen on `HM-Required` as that implies to me that our other standards are "Extra" and not "Required". How about `HM-Minimum`?

also on this topic, w/r/t:

> I think we want to extend this with an "Altis" standard that includes this but overrides a bunch of the messages to point to Altis-specific documentation. We might want to include that here too, but not sure.

I would very much like to keep HM and Altis coding standards configs separate. We have and will for a long while have large clients not on Altis and pushing Altis messaging or notices to those client developers make for dis-congruent messaging.

HM/Sniffs/Performance/SlowMetaQuerySniff.php
```diff
@@ -0,0 +1,315 @@
...
+	 * Check an individual meta_query item.
+	 *
+	 * @param int $array_open Token pointer for the array open token.
+	 */
+	protected function check_meta_query_item( int $array_open ) {
+		$array_open_token = $this->tokens[ $array_open ];
+		if ( $array_open_token['code'] !== T_ARRAY && $array_open_token['code'] !== T_OPEN_SHORT_ARRAY ) {
+			// Dynamic value, we can't check.
+			$this->addMessage(
+				'meta_query is dynamic, cannot be checked.',
```
mikeselander: Is the best course of action here for a developer to add a single-line-excluding comment? The message isn't very clear about what to do in this case. I think flagging these is valuable but I'd like to have a clear path forward.

As I think about this, would this be documented in the Altis docs or our Engineering handbook maybe?

HM/Sniffs/Performance/SlowMetaQuerySniff.php
mikeselander: Also, could we get tests for these custom sniffs please?

HM-Required/ruleset.xml
```diff
@@ -0,0 +1,307 @@
...
+				<element value="json_decode" />
+
+				<!-- wp_trim_words() uses wp_strip_all_tags() internally. -->
+				<element value="wp_trim_words" />
+			</property>
+		</properties>
+	</rule>
+
+	<!-- Require strict types for in_array() calls. -->
+	<!-- <rule ref="WordPress.PHP.StrictInArray"/> -->
```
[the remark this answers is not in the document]
rmccue: It's not super required, in the same way that we don't require `===` for checks. They're best practice, so good for HM, but not for minimum.

HM/Sniffs/Performance/SlowMetaQuerySniff.php
rmccue: It just predates the short array standard; in this case, I was also copying the WPCS style. We can convert them all over in the future if we want?

HM/Sniffs/Performance/SlowMetaQuerySniff.php
rmccue: Written from scratch, planning on contributing back to wpcs.

HM-Required/ruleset.xml
rmccue: Hmm, I think actually that should be fine, will remove

HM/Sniffs/Performance/SlowMetaQuerySniff.php
rmccue: We'll document in both.

Best course of action is to add `// phpcs:ignore HM.Performance.SlowMetaQuery -- some comment to explain why this is OK`

rmccue:
> On slow meta queries, I had `'meta_key'     => 'deploy_id', // WPCS: slow query ok.` does this style of exclude no longer work?

Seems like maybe that has broken. Should switch to new style, I don't know if it's worth supporting the old?

> Getting a few ` 189 | ERROR   | Missing wp_unslash() before sanitization.`, is that actually a security issue? Or is this just a bug/unexpected functionality sniff?

That shouldn't be a false positive; do you have examples of this?

> I'm getting "All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks) | found 'str_repeat'" for `echo ':' . str_repeat( ' ', 2048 ) . "\n\n";` static analysis should be able to see there's no dynamic data here?

Alas, cannot detect this without going deeper I think, unless we specifically allow `str_repeat`

> It's showing errors for the unit test files in `./tests` maybe we should add a default exclude for that?

Yeah, perhaps in the Altis standard.

rmccue:
> twentyseventeen is failing a lot of checks, might be worth checking the default themes too. I feel like we probably shouldn't be failing those as I'd _expect_ it's not the theme that's wrong, its our static analysis.

Added a couple extra excluded functions, and Twenty Twenty now passes just fine. The others don't, and I'm 99% certain they're excluded in core's own ruleset, because they were developed before core used the sniffer.

rmccue:
Merged in the major upgrade from master, and it seems OK. @smccafferty has been working on documenting this at https://github.com/humanmade/altis-documentation/pull/127

Still need to write tests for the custom sniffs.

kadamwhite:
@rmccue to

- rename the code standard
- write those tests for custom sniffs

This needs to happen as part of the current Altis release cycle, so we're allowed to hound Ryan for those tasks until he's done them!

HM/Sniffs/Security/NonceVerificationSniff.php
```diff
@@ -0,0 +1,43 @@
...
+ * This is subclassed from WPCS to allow `$_GET` variables to be used if
+ * configured to do so.
+ */
+class NonceVerificationSniff extends WPCSNonceVerificationSniff {
+	use ExtraSniffCode;
+
+	/**
+	 * Allow query ($_GET) variables to be used without checking nonces?
+	 *
+	 * Nonces are designed to protected against destructive actions taking
```
mikeselander, line 21:
```suggestion
	 * Nonces are designed to protect against destructive actions taking
```

HM/Sniffs/Performance/SlowMetaQuerySniff.php
mikeselander: I'm OK with converting these all at once later 👍

rmccue:
> ```
> ERROR   | Detected usage of a possibly undefined superglobal array index: $_REQUEST['filters']. Use isset() or empty() to check the index exists before using it (HM.Security.ValidatedSanitizedInput.InputNotValidated)
> ```

This one is definitely intentional, as it can cause warnings/etc.

> ```
> WARNING | Using the WPCS native whitelist comments is deprecated. Please use the PHPCS native "phpcs:ignore Standard.Category.SniffName.ErrorCode" annotations instead. Found: // WPCS: XSS OK. (HM.Security.EscapeOutput.DeprecatedWhitelistCommentFound)
> ```

This one is mostly intentional as well (at least, I kept it in intentionally).

> ```
> ERROR | The $text arg must be a single string literal, not "'Enable the Block Editor for this '.$postType".(WordPress.WP.I18n.NonSingularStringLiteralText)
> ```

This one is debatable; I did leave it in given it's a straight-up broken translation, but we could drop it out if we want?

HM-Minimum/ruleset.xml
```diff
@@ -0,0 +1,304 @@
...
+				<element value="self_admin_url" />
+
+				<!-- Other URL functions -->
+				<element value="get_template_directory_uri" />
+				<element value="get_theme_file_uri" />
+				<element value="get_term_link" />
+				<element value="wp_nonce_url" />
+
+				<!--
+					For the minimum standard, we're not treating translations
```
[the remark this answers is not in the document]
rmccue: We can if we want, yeah. Shall I add in this PR?

rmccue:
Docs are now live: https://www.altis-dxp.com/resources/docs/guides/code-review/standards/

Let's ship!

PR #181, opened 2020-03-05
Add better checks for meta queries

rmccue:
This replaces the blanket rule in WPCS for any meta query with something a lot smarter, allowing any meta queries which are using `EXISTS` or `NOT EXISTS` queries.

It also disallows the use of `meta_value`, `meta_value_num`, and `rand` in `orderby`.

### DIFF
--- a/HM/Sniffs/Performance/SlowMetaQuerySniff.php
+++ b/HM/Sniffs/Performance/SlowMetaQuerySniff.php
@@ -143,16 +143,8 @@ class SlowMetaQuerySniff extends AbstractArrayAssignmentRestrictionsSniff {
 				$compare = $default_compare;
 			}
 
-			if ( $compare !== 'EXISTS' && $compare !== 'NOT EXISTS' ) {
-				// Add a message ourselves.
-				$this->addMessage(
-					'meta_query is using %s comparison, which is non-performant.',
-					$this->stackPtr,
-					'warning',
-					'nonperformant_comparison',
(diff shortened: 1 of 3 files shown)


HM/Sniffs/Performance/SlowMetaQuerySniff.php
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	/**
+	 * Recursively check a meta_query value.
+	 */
+	protected function check_meta_query() {
+		// Grab the token we're detecting.
+		$token = $this->tokens[ $this->stackPtr ];
+		if ( $token['code'] !== T_DOUBLE_ARROW ) {
+			var_dump( $token );
```
joehoyle: Is this meant to be here?

rmccue: Whoops, cleaned up.

It never actually gets hit (because we're working off the `T_DOUBLE_ARROW` token), was just there for double-checking that originally.

rmccue:
Yes, it checks any array for those keys; that's how the others in WPCS work already.

tfrommen:
@rmccue this rule flags any meta query using `relation`. Is this a bug (I suppose it is)?

```
'meta_query' => [
    'relation' => 'OR', // <-- this breaks the sniff
    [ ... ],
],
```

Happy to provide a PR to address that, but I wanted to make sure I understood the intention behind this change.

PR #185, opened 2020-03-25
Show colors and rule names in the report

igmoweb:
When running `phpcs`, the summary report looks ugly and doesn't show any information about the rules that are not passing. This commit helps to:
- Use some colors in the terminal
- Display, for every rule that is not passing, the name of the rule to help to identify what's really going on.

Here's a sample:

**Before**
Captura de pantalla 2020-03-25 a las 11 19 49

**After**
Captura de pantalla 2020-03-25 a las 11 19 39

### DIFF
--- a/HM/ruleset.xml
+++ b/HM/ruleset.xml
@@ -7,6 +7,8 @@
 	<exclude-pattern>node_modules/*</exclude-pattern>
 	<exclude-pattern>vendor/*</exclude-pattern>
 
+	<arg name="colors"/>
+	<arg value="s"/>
 	<arg name="basepath" value="." />
 	<arg name="extensions" value="php" />
 


mikeselander:
I'm a big fan of showing the colors 👏 that's an excellent change, hank you for pushing that in! 

I'm not a fan of showing the rules by default as these leads to extremely verbose output and a project can always show the rules for an individual run if needed. I'm also unsure how this would affect Linter Bot behavior...

This is just an opinion though, @joehoyle @rmccue @fklein-lu or @kadamwhite  any thoughts/opinions on passing the `-s` flag by default?

joehoyle:
Hmm I’m not sure why we would want to define either of these tbh, as they are intended to be specified by the caller. If anyone wants colors they should specify the option when running phpcs. If they _always_ want colors they can configure their tooling accordingly. I don’t know to what extent this might mess with editor integrations too that don’t expect color encoded output.

If this is intended for people running the standard via phpcs, I don’t think it’s a good idea to arbitrarily change the phpcs defaults

kadamwhite:
As we can't override "defaults" from the CLI, there'd be no good way to turn these options off if they are not desired -- we should probably document in the README how to set the CLI to pass these flags, but hold off on modifying the config itself to accommodate CI logs and other situations where we would want alternate behavior.

Issue #184, opened 2020-03-17
Standards not being detected by VS Code


johnbillion:
Given VS Code and the phpcs extension (either https://marketplace.visualstudio.com/items?itemName=ikappas.phpcs or the https://marketplace.visualstudio.com/items?itemName=shevaua.phpcs fork), the HM ruleset does not appear to get detected at all. Swapping out the HM ruleset for the WordPress-Extra ruleset and changing nothing else, results in the sniffs working.

Here's an example repo: https://github.com/johnbillion/phpcstest

* Clone the above repo
* Run `composer install`
* Open the repo in VS Code
* Open index.php and observe that the standards violation on line 5 does not get detected
* Open phpcs.xml.dist, comment out the HM line, and uncomment the WordPress-Extra line, and observe that the standards violation will be correctly detected

The `Output` panel does not show any errors.

I have no global dependencies installed.

johnbillion:
Note that running `./vendor/bin/phpcs .` on the CLI works as expected.

ocean90:
I remember [something similar](https://github.com/wearerequired/coding-standards/issues/27) in our standard. Try removing [this](https://github.com/humanmade/coding-standards/blob/0fb7ee17abf804e5adb5a75ee3e71aff453232fa/HM/ruleset.xml#L5) and/or [this](https://github.com/humanmade/coding-standards/blob/0fb7ee17abf804e5adb5a75ee3e71aff453232fa/HM/ruleset.xml#L10) lines.

mikeselander:
hhhmmmm That first line came up as an issue in #159 as well. I think it might be time to remove if it's indeed causing these issues as well.

svandragt:
For future reference.
The workaround presented here breaks PHPStorm's autoconfiguration to run PHPCS as it cannot find WordPress-Extra sniffs. Changing it back to HM fixes that as well.

Issue #159, opened 2019-12-18
current ruleset prevents to exclude everything except certain folders


igmoweb:
[This commit](https://github.com/humanmade/coding-standards/commit/deb4437aee5565a2784aa31bbb585d6de358bcc3
) introduced a new line in the ruleset file:

`<file>.</file>`

And that's preventing to use HM coding standards in projects that needs to ignore everything except certain folders. Apparently, phpcs does respect the HM line instead of the one inside the project xml. Why? Who knows but if I use `<file>path/to/something/</file>` phpcs does not respect it and for some reason it doesn't queue any file to lint.

The target of adding that line was to simplify CLI commands and avoid that annoying `.` when running phpcs commands. I think is better to be annoyed by this that prevent different folder structures in our projects but I'm opening this to create a discussion about different strategies if there are.

More [in the Slack conversation](https://hmn.slack.com/archives/C03K3J34A/p1576665935120200)

PR #187, opened 2020-04-06
Remove Default `<file>` and `<basepath>` values

mikeselander:
This PR fixes several issues related to the defaults set in the HM ruleset that define the basepath and main files. These provide easy defaults, but can cause issues in several cases which make our rules unusable.

 - Removing `<basepath>`: This value was preventing Visual Studio Code PHPCS extensions from using our ruleset. See #184
 - Removing `<file>`: This value was preventing projects from creating an "inclusion" list, forcing them to instead build a "disallow" list instead which is untenable for some projects. See #159

I feel comfortable forcing projects to set these values as they need since PHPCS should really only be setup once on a project.

Paging @jrfnl and @ntwb - do you see any problem with removing these values I might have missed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,16 @@
 ## Unreleased (1.0.0)
 
-### Added
+### Added:
  - Included `WordPress-Docs` by default in PHPCS #177
  - Add ESLint rule for JSX boolean values #183
 
-### Updated
+### Updated:
  - Updated WPCS to 2.2.1 #151
  - Updated VIPCS to 2.0.0 #151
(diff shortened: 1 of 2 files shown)


jrfnl:
@mikeselander No, those are settings which should never have been included in a Standard anyway. They should only ever be used in the rulesets of individual projects.

mikeselander:
> The <config name="testVersion" value="7.1-" /> should also be project based as it is near impossible to overrule otherwise.

Good to know, I'll address that in another PR 👍 

Thanks for the feedback, all - moving this forward :shipit:

PR #188, opened 2020-04-06
Disallow `use function` and `use const`

### LINKED ISSUE #158: Consider disallowing use of `use function...`
In line with https://github.com/humanmade/Engineering/pull/143, I propose that we add a rule dis-allowing the use of `use function`. 

As noted by Joe in https://github.com/humanmade/Engineering/pull/143#issuecomment-560330827, this will break linting for several current projects. I personally feel that this is worth it as the rule can be ignored on current projects and it will enforce properly on all new projects.

mikeselander:
This addition disallows developers from using functions, while keeping the ability to `use` namespaces and constants. 

Special thanks to @jrfnl for building this ruleset for us and making this so easy to use!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,8 @@
 	"description": "Human Made coding standards",
 	"type": "phpcodesniffer-standard",
 	"license": "GPL-2.0-or-later",
+	"minimum-stability": "dev",
+	"prefer-stable": true,
 	"require": {
 		"php": ">=7.1",
 		"wp-coding-standards/wpcs": "2.2.1",
@@ -10,7 +12,8 @@
 		"fig-r/psr2r-sniffer": "^0.5.0",
 		"phpcompatibility/phpcompatibility-wp": "^2.0.0",
 		"squizlabs/php_codesniffer": "~3.5.0",
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.6.0"
+		"dealerdirect/phpcodesniffer-composer-installer": "^0.6.0",
+		"phpcsstandards/phpcsextra": "^1.0"
 	},
 	"require-dev": {
 		"phpunit/phpunit": "^5.7"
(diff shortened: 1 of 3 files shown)


jrfnl:
> Special thanks to @jrfnl for building this ruleset for us and making this so easy to use!

Well, didn't build it _just_ for you, but glad you like it. ;-)

composer.json
```diff
@@ -3,14 +3,17 @@
 	"description": "Human Made coding standards",
 	"type": "phpcodesniffer-standard",
 	"license": "GPL-2.0-or-later",
+	"minimum-stability": "dev",
+	"prefer-stable": true,
```
rmccue, line 7: Are these changes intentional? Neither of them will have any effect if the project isn't the root project.

mikeselander: They were, yes. `phpcsextra` is only released as `alpha` and it wouldn't install without adding these in.

rmccue: OK, we'll have to hold on releasing this then until phpcsextra is released, as the minimum-stability won't flow through, and would make the coding standards uninstallable as a dependency.

roborourke:
Requested a stable release: https://github.com/PHPCSStandards/PHPCSExtra/issues/91

kadamwhite:
@roborourke and cc @rmccue for the legacy piece -- we still want this? Could probably get it in now with a refreshed patch

roborourke:
Unless there are newer releases then yes I think so. Is it still needed even with the latest WPCS versions? 

I can't remember the logic behind disallowing them but I don't care that much personally.

goldenapples:
:+1: to refreshing this so it can be merged. Our [coding standards](https://engineering.hmn.md/standards/style/php/#namespace-and-function-imports) are pretty explicit about this:

> * Importing functions explicitly with `use function` should be avoided.
> * Aliasing functions explicitly with `use function ... as` should also be avoided.
> * Parent namespaces are preferred for importing functions.

and it would be nice to have this enforced by phpcs rather than discussing case by case.

rmccue:
Yep, still an intentional part of the standards so +1 to getting it in

Issue #75, opened 2018-08-07
Enforce props alphabetical sorting


igmoweb:
Finding props sometimes in a React file is a bit difficult due to the random order. We could consider [jsx-sort-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md) rule to automagically sort them alphabetically.

In the same line, we could also think about adding [https://eslint.org/docs/rules/sort-imports](https://eslint.org/docs/rules/sort-imports) to sort imports too.

kadamwhite:
I'm :-1: on requiring alphabetization of component properties. We experimented with it at my last company, and it was a consistent distraction and annoyance for the majority of the team. While the intent is good, in practice it means that you usually can't group related props together, and spend more time (even with `--fix`) thinking about and updating prop order than you do actually writing useful code. It slowed down our component development noticeably to have to keep looking over our shoulder to make sure we weren't doing things wrong.

I'd also argue that this papers over the bigger issue of, if you have a component that includes so many properties you need to order them, then it should probably be split up into different components or you should use containers further down in the tree.

tl;dr, we tried it, tor three projects in a row it got cited as more trouble than it was worth. I'm in favor of alphabetical ordering of imports (Ryan's proposed ("absolute, then relative, then side effects, with a blank line between each'), and I'm fine with this being an optional rule we can add per-project, but I will do what I can to prevent `jsx-sort-props` from being adopted within the formal & official Human Made coding standards :)

kadamwhite:
Broke import sorting out into a separate issue so we can discuss separately.

igmoweb:
@kadamwhite Sounds a good reason to me. Thanks!

mattheu:
I'm re-opening this issue because it is stated in our coding standards that we should have props in alphabetical order. 

Personally I'm 👍 on enforcing the ordering, but i'm not too hung up either way. However I don't think it should really be in the written coding standards if we're not enforcing the rule in our linters. If its optional, may as well remove it! This has already caused some confusion for @BronsonQuick. 

I think [the rules for sorting HTML properties
](https://new-engineering.hmn.md/standards/style/react/#align-and-sort-html-properties) would be covered by [jsx-sort-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md), but we could go further and add [eslint sort keys](https://eslint.org/docs/rules/sort-keys) too.

kadamwhite:
[...]

PR #195, opened 2020-05-08
Add React sorting props rule

igmoweb:
Based on https://github.com/humanmade/coding-standards/issues/75 conversation, this PR adds a new rule to force the React components props to be sorted alphabetically

### DIFF
--- a/packages/eslint-config-humanmade/index.js
+++ b/packages/eslint-config-humanmade/index.js
@@ -105,6 +105,7 @@ module.exports = {
 			'singleline': 'consistent',
 		} ],
 		'react/jsx-boolean-value': [ 'error', 'never' ],
+		'react/jsx-sort-props': [ 'error' ],
 		'jsx-a11y/anchor-is-valid': [ 'error' ],
 	},
 };
--- a/packages/eslint-config-humanmade/index.js
+++ b/packages/eslint-config-humanmade/index.js
@@ -105,7 +105,10 @@ module.exports = {
 			'singleline': 'consistent',
 		} ],
 		'react/jsx-boolean-value': [ 'error', 'never' ],
-		'react/jsx-sort-props': [ 'error' ],
+		'react/jsx-sort-props': [ 'error', {
+			'reservedFirst': [ 'key', 'ref' ],
+			'callbacksLast': true,
+		} ],
 		'jsx-a11y/anchor-is-valid': [ 'error' ],
 	},
 };
--- a/packages/eslint-config-humanmade/index.js
+++ b/packages/eslint-config-humanmade/index.js
@@ -105,7 +105,7 @@ module.exports = {
 			'singleline': 'consistent',
 		} ],
 		'react/jsx-boolean-value': [ 'error', 'never' ],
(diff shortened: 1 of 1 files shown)


rmccue:
Pinging @kadamwhite for review, as I think he was the main dissenter to the original issue; let's make sure we're all happy.

packages/eslint-config-humanmade/index.js
```diff
@@ -105,6 +105,7 @@ module.exports = {
 			'singleline': 'consistent',
 		} ],
 		'react/jsx-boolean-value': [ 'error', 'never' ],
+		'react/jsx-sort-props': [ 'error' ],
```
rmccue, line 108: I think this needs to include `callbacksLast: true` and `reservedFirst: true`

rmccue: Actually I think to align with our handbook, specifically `reservedFirst: ["key"]`.

Should the handbook say to put `ref` or anything else first too? Default for the rule if it's `true` is `key`, `dangerouslySetInnerHTML`, `ref`, and `children`; I feel like I wouldn't want anything more than `key` and `ref` though, as the others aren't really "special"

igmoweb: @rmccue I've fixed the PR by adding those two properties. I've included `ref` as I think that is short enough to miss it when reading the props and keeping it on top highlights it a bit more.

rmccue: To be clear, it's not that they're short, it's that they're not standard props. `key` is used in the reconciler, and isn't passed as an actual prop to the component; `ref` is used during the mounting process, and also isn't passed through (although it's a bit less special).

I think that's also why the others make sense to me to not include in the exclusion; `dangerouslySetInnerHTML` and `children` are fairly regular props.

kadamwhite:
To clarify the above following further discussion with @rmccue the issue I have is that it’s an Error, which interrupts and distracts from more significant issues while developing. If this was a warning, which would be flagged in CI and should be addressed before commit, I’d be in favor of merge.

packages/eslint-config-humanmade/index.js
```diff
@@ -105,6 +105,10 @@ module.exports = {
 			'singleline': 'consistent',
 		} ],
 		'react/jsx-boolean-value': [ 'error', 'never' ],
+		'react/jsx-sort-props': [ 'error', {
```
rmccue, line 108: Per discussion, let's make sure this is a warning. hm-linter treats warnings as errors for builds anyway, so this is mainly to improve the usability locally.

```suggestion
		'react/jsx-sort-props': [ 'warn', {
```

rmccue:
> please don’t ask me to rubber stamp it

FWIW, I do legitimately want your feedback and to get to the heart of your concerns; the local vs CI point I think is definitely valid, and it sounds like switching this to a warning makes us all at least somewhat happy, which I would call a win :)

igmoweb:
@rmccue @kadamwhite I've switched to warnings instead of errors in the code.

PR #198, opened 2020-05-20
Remove Final PHPCS settings

### LINKED ISSUE #191: Re-consider setting `testVersion` value by default
In https://github.com/humanmade/coding-standards/pull/187#issuecomment-610105370 it was identified that we should likely not be setting the `testVersion` value by default as it cannot be updated.

> The `<config name="testVersion" value="7.1-" />` should also be project based as it is near impossible to overrule otherwise.

We should look at removing this for 1.0.0 with the other ruleset-default breaking changes.

mikeselander:
Removing the last of the PHPCS settings that were preventing projects from setting their own values effectively. This makes our ruleset a more effective Standard in that we're no longer making assumptions for end users.

Also added examples of the values that may be set  per-project to the readme.

Resolves #191

### DIFF
--- a/HM/ruleset.xml
+++ b/HM/ruleset.xml
@@ -5,6 +5,8 @@
 	<exclude-pattern>node_modules/*</exclude-pattern>
 	<exclude-pattern>vendor/*</exclude-pattern>
 
+	<arg name="extensions" value="php" />
+
 	<autoload>bootstrap.php</autoload>
 
 	<!-- Check for PHP cross-version compatibility. -->
--- a/HM/ruleset.xml
+++ b/HM/ruleset.xml
@@ -5,12 +5,9 @@
 	<exclude-pattern>node_modules/*</exclude-pattern>
 	<exclude-pattern>vendor/*</exclude-pattern>
 
-	<arg name="extensions" value="php" />
-
 	<autoload>bootstrap.php</autoload>
 
 	<!-- Check for PHP cross-version compatibility. -->
-	<config name="testVersion" value="7.1-" />
 	<rule ref="PHPCompatibilityWP" />
 
 	<rule ref="WordPress-Core">
--- a/readme.md
+++ b/readme.md
@@ -80,17 +80,17 @@ If you want to add further rules (such as WordPress.com VIP-specific rules) or c
 ```xml
 <?xml version="1.0"?>
 <ruleset>
-    <!-- Files or directories to check -->
-    <file>.</file>
+	<!-- Files or directories to check -->
+	<file>.</file>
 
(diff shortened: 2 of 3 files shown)


HM/ruleset.xml
```diff
@@ -5,12 +5,9 @@
 	<exclude-pattern>node_modules/*</exclude-pattern>
 	<exclude-pattern>vendor/*</exclude-pattern>
 
-	<arg name="extensions" value="php" />
```
rmccue, line 8: Ooh, this one I think we might actually want to keep? What's the default in phpcs?

readme.md
```diff
@@ -75,11 +75,23 @@ Patterns are relative to the directory that the `.phpcsignore` file lives in. On
...
+    <file>.</file>
+
+    <!-- Path to strip from the front of file paths inside reports (displays shorter paths) -->
+    <arg name="basepath" value="." />
+
+    <!-- Only parse PHP files -->
+    <arg name="extensions" value="php" />
+
+    <!-- Set a minimum PHP version for PHPCompatibility -->
+    <config name="testVersion" value="7.2-" />
```
rmccue, line 93: Mixed tabs and spaces here.

HM/ruleset.xml
mikeselander: By default it look at PHP, JS, and CSS.

It's a sensible default, but the problem is that this is apparently nearly impossible to override in an individual project which makes it somewhat problematic.

ntwb: I think this should be mentioned in the docs, otherwise without the `<arg name="extensions" value="php" />` line in your config it will check JS and CSS files be default

mikeselander: Really, it's OK to leave this in here if no one has needed to override this value before. We have all the linters to cover the other file types so if there's no need, we don't _have_ to remove it. 

I was aiming to be a more compatible/friendly standard overall, but this is ultimately for us so we can do what we need to make it easiest to work with.

Also, I added some documentation to the readme and we'll have a fairly in-depth post on the Dev P2 once this gets published.

rmccue: Yeah, I'd say we leave it; the JS/CSS stuff is *really* janky, and doesn't really work with modern ES6 and SCSS code properly. I've never seen anyone turn it on, only people wonder why PHP sniffs were weirdly getting applied to JS files too.

(Originally, it was useful when we were running ESLint via phpcs, but the phpcs sniffs often conflicted with our JS standards anyway.)

mikeselander: Done ✅

PR #199, opened 2020-05-20
Adjust Stylelint class and ID selector patterns

### LINKED ISSUE #193: `selector-class-pattern` doesn't match BEM
In the WordPress stylelint standards version 16.0.0, the selector-class-pattern was updated, but it doesn't allow for BEM naming patterns which can cause a huge quantity of stylelint errors.

I propose that we either remove the rule, or add a BEM-specific pattern in our ruleset and allow project to overrule this rule individually. I tried this rule out on a project when testing the upgrade for them and it works nicely in almost all situations: https://github.com/postcss/postcss-bem-linter/issues/82#issuecomment-292464488

mikeselander:
As mentioned in #193, the WordPress `selector-class pattern` does not allow for BEM, and the ID pattern is also unnecessarily restrictive.

This PR:
 - Modifies the `selector-class-pattern` rule to require BEM-type naming. See examples below.
 - Removes the `selector-id-pattern` rule as we have no standard for naming ID.

cc @humanmade/frontend 

Passes:
```css
.something {}
.some-class {}
.anything-else__and--other-things {}
.blocky-mc-block-face {}
.block__element {}
.block__element--modifier {}
.long-block__long-element--long-modifier {}
```

Errors:
```css
.block__element__element {}
.block__element--modifier--modifier {}
.block--modifier--modifier {}
.block--modifier__element {}
```

You can play around with these examples here: https://regex101.com/r/pH8CMO/1

Fixes #193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
  - Add ESLint rule for sorting JSX props #195  
 
 ### Updated:
+ - Adjust Stylelint class and ID selector patterns #199
  - Updated WPCS to 2.2.1 #151
  - Updated VIPCS to 2.0.0 #151
  - Updated DealerDirect to 0.6 #151
--- a/packages/stylelint-config/.stylelintrc.json
+++ b/packages/stylelint-config/.stylelintrc.json
@@ -51,7 +51,7 @@
       }
     ],
     "number-max-precision": 3,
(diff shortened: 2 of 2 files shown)


ocean90:
Thanks!

Here's an alternative version: https://regex101.com/r/pH8CMO/3
It follows the [two dashes style](https://en.bem.info/methodology/naming-convention/#two-dashes-style) and

* doesn't allow to start a class selector with a number
* only allows lowercase characters
* doesn't allow `\\` (Not sure where that is coming from?)

mikeselander:
@ocean90 that is better, thank you!

ocean90:
You might want to enable [`resolveNestedSelectors`](https://stylelint.io/user-guide/rules/selector-class-pattern#resolvenestedselectors-true--false-default-false) as well so nested selectors can be validated as one selector.

mikeselander:
Oh, that is quite neat. Thanks for all your help with this!

PR #206, opened 2020-07-06
Exclude missing package tag file comment rule

### LINKED ISSUE #203: Disable @package requirement
When we added WordPress-Docs, that appears to check for `@package` in the file header. We should turn that off; it's not useful for namespaced code.

Also, we probably want to go over the docs standard with a fine-toothed comb to make sure we know what we're getting.

pdewouters:
Fixes #203


kadamwhite:
No changelog entry needed because this never ended up in a tagged release

PR #208, opened 2020-07-09
Ignore the full stop rules for comments

mikeselander:
The full stop flags in the Docs ruleset is nice for consistency, but generates a huge amount of noise and is generally has not other use. Removing it here to avoid the pain when people install 1.0


mikeselander:
Nah, I think we can skip for the same reason as #206

kadamwhite:
Actually, I've answered my own question because it's the same result as our last exclusion: since we added this ruleset since the last version, shouldn't need to

kadamwhite:
lol jinx. Awesome.

PR #209, opened 2020-07-09
Require JSDoc for functions other than anonymous arrows.

### LINKED ISSUE #157: Consider enforcing use of JSDoc comments
ESlint provides rulesets for enforcing and checking JSDoc comments.
`valid-jsdoc` and `require-jsdoc`

https://eslint.org/docs/rules/valid-jsdoc
https://eslint.org/docs/rules/require-jsdoc

kadamwhite:
Per https://eslint.org/docs/rules/require-jsdoc this set of options appears to implement what was discussed in #157. Anonymous arrows are excluded from JSDoc requirements, while classes, function declarations, and methods all require documentation.

Fixes #157.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Added:
  - Included `WordPress-Docs` by default in PHPCS #177
+ - Add ESLint rule for requiring docblocks #209
  - Add ESLint rule for JSX boolean values #183
  - Add ESLint rule for sorting JSX props #195  
  - Add ESLInt Rules of Hooks ruleset #197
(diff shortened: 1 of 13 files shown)


kadamwhite:
I've added a fixture to prove JSDoc is not required for inline arrow functions, and I've _removed_ the requirement to document class and object methods. While I believe it is usually _very preferable_ to document methods, in React components in particular there isn't a ton of semantic value in putting
```/** @returns {React.ReactNode} The rendered component```
in front of every render method. Or so I believe, anyway. If y'all disagree, please weigh in and we can re-enable `MethodDefinition: true`.

mikeselander:
It _is_ preferable to document vanilla JS methods, it's mostly just React where it would be an issue. Is there a way to turn it off for specific class extends or with a React context only?

kadamwhite:
@mikeselander i unfortunately haven't been able to find one. I can look again, or if you see anything let me know

PR #222, opened 2020-09-07
Support ESLint v6 & v7 in `peerDependencies`

ntwb:
Draft: This is a WIP, adding the multiple ESLint versions to `peerDependencies` may cause issues with the install instructions: `npx install-peerdeps --dev @humanmade/eslint-config@latest` and needs further testing

Right now if installing `@humanmade/eslint-config` v1.0.0 with ESLint v6 or v7 installed the following message is observed:

> `npm WARN @humanmade/eslint-config@1.0.0 requires a peer of eslint@^5.10.0 but none is installed. You must install peer dependencies yourself.`

Adding v6 and v7 to the `peerDependencies` will resolve this

### DIFF
--- a/packages/eslint-config-humanmade/package.json
+++ b/packages/eslint-config-humanmade/package.json
@@ -44,7 +44,7 @@
 	},
 	"peerDependencies": {
 		"babel-eslint": "^10.0.0",
-		"eslint": "^5.10.0 || ^6.0.0 || ^7.0.0"
+		"eslint": "^5.10.0 || ^6.0.0 || ^7.0.0",
 		"eslint-config-react-app": "^3.0.5",
 		"eslint-plugin-flowtype": "^3.2.0",
 		"eslint-plugin-import": "^2.14.0",
--- a/packages/eslint-config-humanmade/package.json
+++ b/packages/eslint-config-humanmade/package.json
@@ -44,7 +44,7 @@
 	},
 	"peerDependencies": {
 		"babel-eslint": "^10.0.0",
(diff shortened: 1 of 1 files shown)


kadamwhite:
@ntwb How're you feeling about this? Have you had the opportunity to test any further?

ntwb:
I think this should be fine, we do the same for stylelint configs, eg:

https://github.com/stylelint/stylelint-config-recommended/blob/c5968e63cd2d8a0e7c293cc3fdf4cf85e36ba868/package.json#L35

https://github.com/WordPress-Coding-Standards/stylelint-config-wordpress/blob/7816e565b44cdb16969297f967343f6c8b092be6/package.json#L57

tfrommen:
@ntwb I was about to create an issue for just that (for both ESLint and stylelint, actually).

One thought that I had, though: Instead of adding a list of supported versions such as `^4.0 || ^5.0 || ^6.0`, do we want to specify a minimum version istead, for example, `>= 4.0.0`?
One obvious reason for this is that the former does not work with any new major version, or it would require us to follow up with patch version releases ourselves, just to add more support.
On caveat, on the other hand, is that we would state to support **any** upcoming version, which might not be true.

What do we feel is the right thing to do?

// cc @kadamwhite

pdewouters (approved):

tomjn:
@kadamwhite do you foresee any blockers to this? I'm keen to get this moving :)

kadamwhite (approved):

kadamwhite:
@tomjn no blockers :)

Issue #142, opened 2019-05-23
Add rule to block use of isset with multiple arguments


rmccue:
IMO, `isset( $a, $b )` is unclear, and there's no good reason to use it instead of the more explicit `isset( $a ) && isset( $b )`

Anyone have strong thoughts either way on this?

pdewouters:
I don't see it used that often, I don't think most people even know you can do that. But I don't think it unclear.

fklein-lu:
I do prefer the `isset( $a ) && isset( $b )` version, as it contains the logical operators, which `isset( $a, $b )` does not. Or said differently the first version is clear, even if your understanding of PHP is moderate, the second one requires you to understand what this particular function does in detail.

tfrommen:
Closed as per https://github.com/humanmade/coding-standards/commit/6ad1fc9f5bd57b0f9fa2658e5e9134168fa9b60d.

PR #236, opened 2020-10-27
#142 - Add new Isset sniff

tfrommen:
This PR adds a new sniff, `HM.PHP.Isset` that throws an error for `isset` calls with more than one argument, as per request in #142.

I wasn't sure if wanted to make this an error, or just a warning. Looking at existing warnings and errors, I think we are rather inconsistent. Some Layout and opinionated sniffs throw errors, while others just warn about stuff that makes perfect sense to follow.

Thoughts?

Do we want to make this a part of `HM-Minimum`?


rmccue:
> I wasn't sure if wanted to make this an error, or just a warning.

Right now, they're treated the same by most of our tooling, so it doesn't really matter. Generally though I'd say stylistic stuff would be a warning, if we do ever want to fix that in the future.

> Do we want to make this a part of `HM-Minimum`?

No, as it's not required to build a secure/performant site.

tfrommen:
@rmccue @kadamwhite I just changed this to add a waning now, not an error.

Is this good to go in? 🙂

fklein-lu (approved):

Issue #154, opened 2019-10-16
Add rule to avoid redundant ternaries


rmccue:
I've seen in a few places we have code like this:

```php
return $foo ? true : false;
```

The ternary here is redundant and is actually just casting the variable to a boolean. Using an actual cast is cleaner:

```php
return (bool) $foo;
```

(The reverse is true as well, `return $foo ? false : true;` can be replaced with `return ! $foo`)

kadamwhite:
Is there a PHPCS sniff for this? @rmccue to investigate. I'm :+1: on the change proposed here

kadamwhite:
Neither PHPCS nor WP have a sniff for this.

rmccue:
Also nothing in the wider market :( Let's move to backlog; it's nice, but would require some custom work, so someday...

PR #237, opened 2020-10-28
#154 - Add new Ternary sniff

tfrommen:
This PR adds a new sniff, `HM.PHP.Ternary` that adds a warning for unnecessary ternary expression, as mentioned in #154:
* `$expr ? true : false`;
* `$expr ? false : true`.

See fixtures for (passing and failing) example code.

Output is as follows:

```
  3 | WARNING | Unnecessary ternary found: Instead of "$expr ? true : false", use "(bool) $expr"
  4 | WARNING | Unnecessary ternary found: Instead of "$expr ? false : true", use "! $expr"
```

Can anyone think of any edge-case usage that the sniff would either flag as false-positive, or that it would miss?

I'm not 100% sure I considered all tokens that would _end_ the ternary, which currently are: semicolon, closing parenthesis, closing curly brace (function or any other scope), and also comma (which I just added in a subsequent commit).
Anything else?
Maybe @jrfnl? 🙂

### DIFF
--- a/HM/Sniffs/PHP/TernarySniff.php
+++ b/HM/Sniffs/PHP/TernarySniff.php
@@ -43,11 +43,11 @@ class TernarySniff implements Sniff {
 		}
 
 		$warning = 'Unnecessary ternary found: Instead of "$expr ? %s : %s", use "%s"';
-		$data = [
+		$data = array(
 			$value_if_true_tokens[0]['content'],
 			$value_if_false_tokens[0]['content'],
 			$value_if_true_tokens[0]['content'] === 'true' ? '(bool) $expr' : '! $expr'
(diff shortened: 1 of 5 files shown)


tfrommen:
Instead of a custom sniff, we might want to use the [`UselessTernaryOperatorSniff`](https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/ControlStructures/UselessTernaryOperatorSniff.php) one included in the [Slevomat Coding Standards](https://github.com/slevomat/coding-standard)...? This one is also fixable, which is nice.

I think I will create a separate issue to discuss pulling in the Slevomat Coding Standards...

PR #239, opened 2020-12-10
Exclude doc param alignment rules

### LINKED ISSUE #238: Remove requirement to align docs
In #167 we added sniffs for inline documentation (phpDoc).

In the process, we appear to have also added a sniff that requires alignment of arguments within phpDoc with spaces.

For the same reasons as outlined in #47 (array alignment), we should not be checking against this. Primarily, alignment significantly increases code churn while not improving legibility immensely. This is particularly true for phpDoc which is often parsed and displayed in UI instead, allowing the UI to establish alignment per user preference, so not gaining us really anything from aligning in the code.

We should remove the alignment requirement in documentation.

roborourke:
This fixes #238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.1.2 (December, 2020)
+
+### Removed:
+
+- Disabled requirement to align PHPDoc parameters, inherited from WordPress-Docs in July's 1.0 release #239
+
 ## 1.1.0 (September 18, 2020)
 
 ### Added:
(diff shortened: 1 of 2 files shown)


rmccue (approved):
I am +1 on this, but I'll leave the decision for @kadamwhite (noted in Slack that he approved in principle)

kadamwhite (approved):

kadamwhite:
Travis seems hung up, but the last commit is markdown only. Mergin'.

PR #241, opened 2020-12-10
Retroactively update changelog

kadamwhite:
Realized we needed to backfill the 1.1.1 release...

Also fills in the date for 1.1.2 assuming we're releasing that relaxation of phpdoc alignment today, and adds a composer test script alias to make the testing path discoverable via avenues other than contributing.md


roborourke (approved):
Thanks @kadamwhite !

Issue #245, opened 2021-02-10
Update and review usage of the PSR-2-R PHP coding standard


tfrommen:
The HM Coding Standards make use of the external [PSR-2-R](https://github.com/php-fig-rectified/psr2r-sniffer) standard. There are some issues with this.

Currently, we are using the following three rules as part of the HM standard:

- `PSR2R.ControlStructures.NoInlineAssignment` 
- `PSR2R.Namespaces.UnusedUseStatement`
- `PSR2R.Namespaces.UseInAlphabeticalOrder` 

Our current version constraint for this standard is `^0.5.0`. Since this is a non-stable version (i.e., below 1.0.0), the caret notation will be treated differently, and actually only allow for patch-level updates. This means anything up to, but not including 0.6.0.

Now, there have been quite a few updates—to be precise: **7**—that we will never see with our current constraint. Some of these releases include breaking changes, and the current version is (stable) [1.0.1](https://github.com/php-fig-rectified/psr2r-sniffer/releases/tag/1.0.1).

One of these breaking changes was to [remove the `PSR2R.Namespaces.UseInAlphabeticalOrder` rule](https://github.com/php-fig-rectified/psr2r-sniffer/commit/3f602977f4eb9cb1c2bfadeef0f0131ee15986fb) as it was conflicting with other tooling such as IDE (as it was technically incorrectly implemented). We should **remove this rule from our standard**.

Then, I was wondering why we are currently only using these three select rules. Overall, [the PSR-2-R standard looks pretty solid](https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide.md) to me, and should be very much in line with everything we already do or want to do.

I see the following items to discuss and action:

- [ ] Remove the `PSR2R.Namespaces.UseInAlphabeticalOrder` rule from the HM standard.
- [ ] Review the PSR-2-R standard for any breaking changes in the rules we use.
- [ ] Update the PSR-2-R standard by specifying the `^1.0` version constraint.
- [ ] Address or document all breaking changes, if any.
- [ ] Consider adding more/all of the PSR-2-R standard to the `HM` standard.
- [ ] Consider adding select PSR-2-R rules to `HM-Minimum`, where this makes sense and is in line with what this standard is about (e.g., security, performance, error prevention, but **no** formatting etc.).

Any thoughts on any of the above?

PR #246, opened 2021-02-10
Remove deprecated/removed PSR-2-R rule

tfrommen:
The `PSR2R.Namespaces.UseInAlphabeticalOrder` rule has been [removed](https://github.com/php-fig-rectified/psr2r-sniffer/commit/3f602977f4eb9cb1c2bfadeef0f0131ee15986fb) from the [PSR-2-R](https://github.com/php-fig-rectified/psr2r-sniffer) standard, so we should remove it, too. As soon as we'd update the standard, it would break anyway. See #245.

### DIFF
--- a/HM/ruleset.xml
+++ b/HM/ruleset.xml
@@ -135,7 +135,6 @@
 		<exclude name="PSR2.Namespaces.UseDeclaration.MultipleDeclarations" />
 	</rule>
 	<rule ref="PSR2R.Namespaces.UnusedUseStatement" />
-	<rule ref="PSR2R.Namespaces.UseInAlphabeticalOrder" />
 
 	<!-- Ban inline assignment in control structures (see note on Yoda Conditions above). -->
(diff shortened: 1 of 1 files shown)


rmccue:
We'll need to copy the rule itself over I think?

tfrommen:
@rmccue well, it's not working as intended, that's why it has been removed. If we want to keep the behavior, I suggest to use the [`AlphabeticallySortedUses` sniff in the Slevomat Coding Standards](https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/Namespaces/AlphabeticallySortedUsesSniff.php). This is already a dependency of PSR-2-R, [as of version 1.0.0](https://github.com/php-fig-rectified/psr2r-sniffer/blob/1.0.0/composer.json#L17).

rmccue:
Works for me; can you update this PR to do so?

Issue #231, opened 2020-10-21
Replace deprecated declaration-property-unit-* rules


tfrommen:
The `declaration-property-unit-blacklist` and the `declaration-property-unit-whitelist` rules have been deprecated, and should be replaced with `declaration-property-unit-disallowed-list` and `declaration-property-unit-allowed-list`, respectively.

This will be a **breaking** change, as user overrides won't work any longer, and therefore require a **major** version release.

**References:**
* https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-unit-blacklist/README.md
* https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-unit-whitelist/README.md

tfrommen:
This is actioned in #248.

ntwb:
This is now fixed via #248 & #275

PR #247, opened 2021-02-23
Fix deprecated rule warnings, update dependencies, commit lockfile.

mattheu:
Running this locally I was hitting some warnings due to a rule having been deprecated. 

Needs bot stylelint and wordpress stylelint rules to be updated as well.


tfrommen:
This is captured in #231. The "problem" that I see with this change is that it's a breaking change. If we want to do this, then it needs to be a major version release.
Other than that, the change makes sense, of course. 🙂

mattheu:
Good point and happy to hold off on it just eager to get rid of the notices :)

tfrommen:
Looking at #248, I guess this one here can now be closed, @ntwb? 🙂

tfrommen:
Closing in favor of #248.

PR #248, opened 2021-02-23
Update `@humanmade/stylelint-config` in preparation for v2 release

mattheu:
I am having some issues with HM Linter bot flagging style issues, but I'm not getting the same errors locally. and I'm a little stumped by it. I don't know for sure, but i think committing the package lock file could help ensure this isn't caused by mismatched versions. 

The actual problem is this: https://github.com/humanmade/HM-Playbook/pull/200, and the errors flagged are related to the rule `function-calc-no-invalid`. Yet its not happening locally. Looks like this was a bug in stylelint when used with sass files, and it was [fixed a year ago](https://github.com/stylelint/stylelint/pull/4046). This makes me suspect that this could be caused by a version mismatch, and therefore may be fixed by committing the lockfile.

### DIFF
--- a/packages/stylelint-config/.stylelintrc.json
+++ b/packages/stylelint-config/.stylelintrc.json
@@ -14,10 +14,10 @@
     "declaration-block-semicolon-newline-after": "always-multi-line",
     "declaration-block-single-line-max-declarations": 4,
     "declaration-no-important": true,
-    "declaration-property-unit-blacklist": {
+    "declaration-property-unit-disallowed-list": {
       "font-size": [ "px" ]
     },
-    "declaration-property-unit-whitelist": {
+    "declaration-property-unit-allowed-list": {
       "animation": [ "ms" ],
       "animation-delay": [ "ms" ],
       "animation-duration": [ "ms" ],
--- a/packages/stylelint-config/.stylelintrc.json
+++ b/packages/stylelint-config/.stylelintrc.json
@@ -1,6 +1,6 @@
 {
   "extends": [
-    "stylelint-config-wordpress/scss"
+    "@wordpress/stylelint-config/scss"
   ],
   "rules": {
     "comment-empty-line-before": [
--- a/packages/stylelint-config/package-lock.json
+++ b/packages/stylelint-config/package-lock.json
@@ -14,9 +14,9 @@
 			}
 		},
 		"@babel/compat-data": {
-			"version": "7.13.5",
-			"resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.13.5.tgz",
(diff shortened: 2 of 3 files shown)


mattheu:
Looking a bit closer... this doesn't necessarily solve my problem either. It looks like the bugfix I'm after was released in [v13.30](https://stylelint.io/changelog#1330), so if HM Linter Bot is using the version of stylelint from here... its still not new enough. 

Either way, I wonder why we're not committing the lockfile. [We do for eslint.](ds/blob/master/packages/eslint-config-humanmade/package-lock.json)

kadamwhite:
@mattheu There's no reason not to include the lockfile that I can see. If you can PR the update / version pin, I can get that pushed out as a patch release

kadamwhite:
@mattheu We include the package lockfile in ESLint because I deployed with NPM, and since Ryan used Yarn, only Yarn was ignored :)

mattheu:
Updated this PR to also update the the stylelint and WP stylelint rules to latest versions.

ntwb:
For the most part this is working as expected testing locally, as @tfrommen highlights [here in Slack](https://hmn.slack.com/archives/C8DQSQFM4/p1614235225025000) that this will need to be shipped as a major SemVer release in HM CS v2.0.

Whilst testing this locally works fine for me, hmlinter linter-bot continues to fail, I need to look closer at either the linter-bot stylelint integration in https://github.com/humanmade/linter-bot/issues/144 or in stylelint itself to resolve the underlying issue.

I've published the updated config here in pull request as it currently stands as `stylelint-next` and is now available to projects to use and test, that is if of course if stylelint is working with linter-bot for your project, to test this PR on your project you can use the following config option for stylelint in `hmlinter.yml` of your project:

```yml
stylelint:
  enabled: true
  version: next
```

mattheu:
Oh now that is a neat idea. Although running this I get the following error: 

`Could not run: Error: Could not find "stylelint-config-wordpress/scss". Do you need a configBasedir?`

ntwb:
> Oh now that is a neat idea. Although running this I get the following error: 
> 
> `Could not run: Error: Could not find "stylelint-config-wordpress/scss". Do you need a configBasedir?`

This was fixed with this commit @mattheu That was fixed in [`8bc47f4` (#248)](https://github.com/humanmade/coding-standards/pull/248/commits/8bc47f4799c6039188c51e25c37e6bc92e08de59)

ntwb:
I've got a few more things to do, will follow this up in further PRs

For now, this can be merged and I'll push up a new `next` release

tfrommen (approved):
If we treat `master` as _home_ for the upcoming v2 releases, this looks alright to me. 👍

PR #269, opened 2021-08-13
Remove deprecated "declaration-property-" rules

alwaysblank:
Both `declaration-property-unit-blacklist` and `declaration-property-unit-whitelist` have been deprecated. The current
ruleset will still run, it will throw deprecation warnings about these two rules. This updates them.

- https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-unit-blacklist/README.md
- https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-unit-whitelist/README.md


ntwb:
Thanks @alwaysblank, closing this in favor of #248, I'm just updating this PR now and should be ready to merge shortly as the upstream stylelint config requires updating for this change in deprecated rule

PR #271, opened 2021-09-08
Ignore "use" within "class" in OrderSniff

kadamwhite:
The error which is reported:

    (
    	[message] => use found on line 48, but require was declared on line 7. Statements should be ordered `namespace`, `use`, `const`, `require`, then code.
    	[source] => HM.Layout.Order.WrongOrder
    	[listener] => HM\Sniffs\Layout\OrderSniff
    	[severity] => 5
    	[fixable] =>
    )

It should be possible to `use` a trait within a class, without triggering the warning for a top-level namespace `use` statement.

### DIFF
--- a/HM/Sniffs/Layout/OrderSniff.php
+++ b/HM/Sniffs/Layout/OrderSniff.php
@@ -51,6 +51,9 @@ class OrderSniff implements Sniff {
 			if ( $next_token['code'] === T_USE && $phpcsFile->findPrevious( T_CLOSURE, $next_pos, null, false, null, true ) !== false ) {
 				continue;
 			}
+			if ( $next_token['code'] === T_USE && $phpcsFile->findPrevious( T_CLASS, $next_pos, null, false, null, true ) !== false ) {
+				continue;
+			}
 
 			// Must be current or higher.
 			$next_type_score = $look_for[ $next_token['code'] ];
--- a/tests/fixtures/pass/use-order.php
+++ b/tests/fixtures/pass/use-order.php
@@ -2,6 +2,7 @@
 
 use Foo\Bar;
 use Foo\Baz as Zztop;
+use Some_Trait;
 
 require( 'some/file/that/exists.php' );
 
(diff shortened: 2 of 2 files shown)


ntwb (approved):

tfrommen:
I agree that the sniff should not take traits into account.

However, the suggested change would not work in all cases, but only if the `use` statement is the very first token in the class body, which also means that it would only work for the first occurrence as well.

All of the following class declarations will (still) be flagged by the Sniff:

```php
class Your_Class {
        const MY_CONST = 42;
        use Some_Trait;
}
```

```php
class His_Class {
        public $my_field;
        use Some_Trait;
}
```

```php
class Her_Class {
        // Some comment.
        use Some_Trait;
}
```

```php
class Their_Class {
        use Some_Trait;
        use Other_Trait;
}
```

When following [PSR12](https://www.php-fig.org/psr/psr-12/) (see _4.2 Using traits_), the `use` statement for traits MUST be the very first **function code** line after the opening curly brace of a class declaration. But that's just for PSR12.
So the question is, should this Sniff adhere to PSR12, or not?

But then again, comments are still an issue, and there could be several flavors of comments. Also, subsequent `use` statements, which are required when following PSR12, are an issue as well.

Maybe we first need to agree on the scope, as well as any assumptions of this sniff before we can adapt it?
For example, maybe we shouldn't try and add exceptions for various `use` statements (or others), but instead prevent them from being checked in the first place...?
To me, it looks like the `HM.Layout.Order` sniff should really only be concerned about the global (file) scope, and not anything nested in that such as classes or namespaces etc.
Do you agree @rmccue?

kadamwhite:
@tfrommen I agree generally about restricting this to top-level. I do think I want to get this specific fix out quickly, to fix things that are broken -- we can then completely remove checking for nested tokens _or_ add specific rules for ordering within classes (making it laxer or stricter, respectively) as a follow-up, without continuing to break validation of existing code

Issue #274, opened 2022-02-17
Package deprecations


tomjn:
```
npm WARN deprecated @stylelint/postcss-markdown@0.36.2: Use the original unforked package instead: postcss-markdown
npm WARN deprecated stylelint-config-wordpress@16.0.0: This package has been deprecated, please use @wordpress/stylelint-config or @wordpress/scripts
```

There may be others, I believe we can fix this by moving from v18 to v20 of `@wordpress/stylelint-config`:

https://github.com/humanmade/coding-standards/blob/master/packages/stylelint-config/package.json#L30

This should eliminate some deprecated package warnings in dependent projects

ntwb:
This should now be resolved with #275 merged, closing this, new version will be published shortly

PR #275, opened 2022-02-17
Update stylelint and the official WP package

tomjn:
This contribute towards #274 by upgrading the official WP package used for configuration and upgrading stylelint to 14.2 which was required by the WP package update. This fixes some deprecation notices that show up downstream in projects

### DIFF
--- a/packages/stylelint-config/package-lock.json
+++ b/packages/stylelint-config/package-lock.json
@@ -1,350 +1,2062 @@
 {
 	"name": "@humanmade/stylelint-config",
 	"version": "1.1.2",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
 	"requires": true,
-	"dependencies": {
-		"@babel/code-frame": {
-			"version": "7.12.13",
-			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
-			"integrity": "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==",
-			"dev": true,
-			"requires": {
-				"@babel/highlight": "^7.12.13"
+	"packages": {
+		"": {
+			"name": "@humanmade/stylelint-config",
+			"version": "1.1.2",
+			"license": "GPL-2.0-or-later",
+			"dependencies": {
+				"@wordpress/stylelint-config": "^20.0.0"
+			},
+			"engines": {
+				"node": ">=12"
+			},
+			"peerDependencies": {
+				"stylelint": "^14.2.0"
 			}
 		},
-		"@babel/compat-data": {
-			"version": "7.13.6",
-			"resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.13.6.tgz",
-			"integrity": "sha512-VhgqKOWYVm7lQXlvbJnWOzwfAQATd2nV52koT0HZ/LdDH0m4DUDwkKYsH+IwpXb+bKPyBJzawA4I6nBKqZcpQw==",
-			"dev": true
-		},
-		"@babel/core": {
-			"version": "7.13.1",
-			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.13.1.tgz",
-			"integrity": "sha512-FzeKfFBG2rmFtGiiMdXZPFt/5R5DXubVi82uYhjGX4Msf+pgYQMCFIqFXZWs5vbIYbf14VeBIgdGI03CDOOM1w==",
-			"dev": true,
-			"requires": {
-				"@babel/code-frame": "^7.12.13",
-				"@babel/generator": "^7.13.0",
-				"@babel/helper-compilation-targets": "^7.13.0",
-				"@babel/helper-module-transforms": "^7.13.0",
-				"@babel/helpers": "^7.13.0",
-				"@babel/parser": "^7.13.0",
-				"@babel/template": "^7.12.13",
-				"@babel/traverse": "^7.13.0",
-				"@babel/types": "^7.13.0",
-				"convert-source-map": "^1.7.0",
-				"debug": "^4.1.0",
-				"gensync": "^1.0.0-beta.2",
-				"json5": "^2.1.2",
-				"lodash": "^4.17.19",
-				"semver": "7.0.0",
-				"source-map": "^0.5.0"
+		"node_modules/@babel/code-frame": {
+			"version": "7.16.7",
+			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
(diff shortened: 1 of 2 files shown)


packages/stylelint-config/package-lock.json
```diff
@@ -1,350 +1,2062 @@
 {
 	"name": "@humanmade/stylelint-config",
 	"version": "1.1.2",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
```
tfrommen: Could you please make sure to use an npm version locally that will produce a **v1 lock file**?

I am all for updating/migrating to more modern tooling (versions), but unless this is required by the stylelint config update, I'd rather do this in a separate, dedicated PR.

packages/stylelint-config/package.json
```diff
@@ -27,10 +27,10 @@
 	},
 	"main": ".stylelintrc.json",
 	"dependencies": {
-		"@wordpress/stylelint-config": "^18.0.0"
+		"@wordpress/stylelint-config": "^20.0.0"
```
tfrommen, line 30: Did you check what kinds of rule changes the jump from 18.x to 20.x means?

Anything new we might not want? Anything removed we currently rely on?

tomjn: the goal of the PR was to start the ball rolling, it was meant to be a draft

packages/stylelint-config/package-lock.json
tfrommen: I don't think there is a required/supported npm version specified anywhere. I just meant that, due to the lock file version change, this is a massive change, and one that will affect people on older versions of npm.

tomjn: it's v3 that's not backwards compatible, and that's only found inside a `node_modules` folder when packages are installed using Node v7+, it should never find its way on to `npmjs.org`

ntwb: The Altis build server pipeline default is Node.js v12 & npm v6

Thus, I'm with @tfrommen here, we should be supporting these versions on projects as the minimum requirements, we can look to upgrade Node.js & npm version support later

packages/stylelint-config/package.json
ntwb: Via https://github.com/WordPress/gutenberg/blob/trunk/packages/stylelint-config/CHANGELOG.md

```md
## 20.0.0 (2022-01-27)

### Breaking Change

-   Increased minimum peer dependency of `stylelint` to `14.2.0` ([#38091](https://github.com/WordPress/gutenberg/pull/38091)). See [official migration guide to v14](https://github.com/stylelint/stylelint/blob/14.0.0/docs/migration-guide/to-14.md) for details.

## 19.0.0 (2021-01-21)

### Breaking Change

-   Increase the minimum Node.js version to 12 ([#27934](https://github.com/WordPress/gutenberg/pull/27934)).
-   Increased minimum peer dependency of `stylelint` to `13.7.0`.

### Bug Fixes

-   Fixed deprecation warning for `declaration-property-unit-whitelist`.
```

packages/stylelint-config/package-lock.json
mattheu: [I've opened a PR to regenerate the lockfile version separately.](https://github.com/humanmade/coding-standards/pull/287) I see Thorsten's point that it makes this PR a big change. 

I think there is no issue doing this. As Stephen says, we're only needing to support v12 for Altis build pipeline and lockfile v2 is backwards compatible.

mattheu:
> there are changes necessary in addition to these, specifically handling of SASS/SCSS changed in Stylelint 14, but I forgot to mark it as a draft and can't find how to change it back

I'd love to push this PR forwards. What are these changes? [Stylelint docs say that if you're wanting to lint scss files you need the rules in a separate package](https://stylelint.io/migration-guide/to-14/#syntax-option-and-automatic-inferral-of-syntax). But we're actually extending the WordPress stylelint config, [which already handles this](https://github.com/WordPress/gutenberg/blob/b17ee9e4835fc561e085bd37a82260f38348109d/packages/stylelint-config/scss.js#L4). So all is fine I think.

packages/stylelint-config/package-lock.json
tomjn: Just reiterating that lock file v2 is and always has been compatible with Node 12 and NPM 6, the NPM docs say so, this is not a major change. The main change in NPM 7 was the introduction of the lockfile version as a concept, not a new lockfile format.

tomjn:
> I'd love to push this PR forwards. What are these changes? [Stylelint docs say that if you're wanting to lint scss files you need the rules in a separate package](https://stylelint.io/migration-guide/to-14/#syntax-option-and-automatic-inferral-of-syntax). But we're actually extending the WordPress stylelint config, [which already handles this](https://github.com/WordPress/gutenberg/blob/b17ee9e4835fc561e085bd37a82260f38348109d/packages/stylelint-config/scss.js#L4). So all is fine I think.

@mattheu if you checkout the branch you'll see there's more to it than just updating the package, configs need accounting for as well as several unknowns I never reached because of that. In moving to a newer major version Stylelint made config changes as well as rule changes that I hadn't anticipated and didn't have time to investigate. If core have resolved those and we can upgrade by upgrading a dependency in WP then that's good but I'm unsure what additional changes are necessary here

mattheu:
I know this was a while back so understand that you don't remember the details of what problems you hit, so i'll try and reproduce to see. 

I checked out this branch and used it to lint the SCSS from a large project currently using `@humanmade/stylelint-config` v1.1.1. I'll summarise the issues below. 

* The custom config on that project used the rules `declaration-property-unit-blacklist` and `declaration-property-unit-whitelist` which are now removed and should be replaced with `declaration-property-unit-disallowed-list` and `declaration-property-unit-allowed-list`.
* Some new rules now causing it to fail
    * No longer allowed to use scss global functions. e.g. switch from `str-index` to `string.index`
    * No longer allowed to use @extend with classes e.g. `@extend .my-component` and should prefer using extend placeholders.
    * No longer alllowed to use the preceeding underscore when importing a partial. e.g. `@import "_partial` should now be `@import "partial" as sass handles this OK. Similarly no need to specify the extension so `@import partial.scss` is also now flagged.  

In my opinion, these changes are all fine and to be expected when updating to a new major release of the linting rules. We should ensure this is rolled out in a major release of our rules too, and document that there are some changes, but otherwise I see no problem with this.

ntwb:
Via https://github.com/WordPress/gutenberg/blob/trunk/packages/stylelint-config/CHANGELOG.md

```md
## 21.0.0 (2022-08-24)

### Breaking Change

    Increase the minimum Node.js version to 14 (https://github.com/WordPress/gutenberg/pull/43141).

## 20.0.0 (2022-01-27)

### Breaking Change

-   Increased minimum peer dependency of `stylelint` to `14.2.0` ([#38091](https://github.com/WordPress/gutenberg/pull/38091)). See [official migration guide to v14](https://github.com/stylelint/stylelint/blob/14.0.0/docs/migration-guide/to-14.md) for details.

## 19.0.0 (2021-01-21)

### Breaking Change

-   Increase the minimum Node.js version to 12 ([#27934](https://github.com/WordPress/gutenberg/pull/27934)).
-   Increased minimum peer dependency of `stylelint` to `13.7.0`.

### Bug Fixes

-   Fixed deprecation warning for `declaration-property-unit-whitelist`.
```

ntwb (approved):
Thanks everyone, apologies it took a little while to get this over the line

PR #282, opened 2022-06-22
Update PHPCS to support PHP 8+

johnbillion:
This change means later versions of PHPCS 3.x can be used, rather than just 3.5. Version 3.7 is required for the sniffs to run on PHP 8.1 without triggering deprecated warnings.

In addition version 3.6 added support for PHP 8 languages features, and 3.7 added support for 8.1 language features.

WPCS and PHPUnit both need to be updated so the tests run on later PHP versions without errors. PHPUnit 7 needs to be used because the base test class in PHPCS that's used for the tests in this library doesn't support PHPUnit 8 ([info here](https://github.com/squizlabs/PHP_CodeSniffer/issues/3490)), hence the use of the ` --ignore-platform-reqs` and `--no-configuration` flags when installing and running the tests on PHP 8 and 8.1.

Todo:

* [x] Determine why the `tests/AllSniffs.php` tests are performing no assertions under PHPUnit 7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,18 @@ jobs:
         - composer install
       script:
         - vendor/bin/phpunit
+    - language: php
+      php: 8.0
+      install:
+        - composer install
+      script:
+        - vendor/bin/phpunit
+    - language: php
+      php: 8.1
+      install:
+        - composer install
+      script:
+        - vendor/bin/phpunit
     - language: node_js
       node_js: 10
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,15 +34,16 @@ jobs:
     - language: php
       php: 8.0
       install:
-        - composer install
+        - composer install --ignore-platform-reqs
       script:
         - vendor/bin/phpunit
     - language: php
       php: 8.1
       install:
-        - composer install
+        - composer install --ignore-platform-reqs
       script:
-        - vendor/bin/phpunit
+        - vendor/bin/phpunit --no-configuration --bootstrap=tests/bootstrap.php tests/FixtureTests.php
+        - vendor/bin/phpunit --no-configuration --bootstrap=tests/bootstrap.php tests/AllSniffs.php
     - language: node_js
       node_js: 10
       install:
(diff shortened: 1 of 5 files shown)


tfrommen:
Right, so, I think this is ready (for use), @johnbillion!

SOmething changed in between PHPUnit 5 and PHPUnit 7 around the definition of "risky" tests. So, the tests are being executed and would fail if something was wrong. But they don't tell PHPUnit what they are doing.

In the PHP_CodeSniffer codebase, they are simply ignoring reports on "useless" test (see [here](https://github.com/squizlabs/PHP_CodeSniffer/commit/9f6385eb32a5f4625e01c54d06a1f9d66201eec2)), which is why they don't see any risky tests. (If you were to pull down PHPCS, install and then run the tests without the `--dont-report-useless-test`, you would see lots of risky tests as well.)

In my mind, the [testSniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/b860f758b87bf43b7e5a7e0ebe1804f2f9771f1b/tests/Standards/AbstractSniffUnitTest.php#L116) method should do something like this:

```php
$assertions = count( $GLOBALS['PHP_CODESNIFFER_SNIFF_CODES'] );

// ...

$this->addToAssertionCount(
	count( $GLOBALS['PHP_CODESNIFFER_SNIFF_CODES'] ) - $assertions
);
```

But given this needs changing in PHP_CodeSniffer, I just followed suit, and silenced useless test reports for now.

I might open an issue/PR for that, when I find time...

svandragt (approved):
Don't see any problems. (reviewed the code, didn't test)

johnbillion:
> So, the tests are being executed and would fail if something was wrong. But they don't tell PHPUnit what they are doing.

Developer: What are you doing?
PHPCS: Nothing.
Developer: Are you sure?
PHPCS: Yep.
Developer: It looks like you're doing something.
PHPCS: Nope.
Developer: Is my code ok?
PHPCS: I cannot possibly comment on the matter.
Developer: Can you tell me if my code is ok?
PHPCS: No I cannot.
Developer: Alright I'm going to make this change which I'm sure is fine and doesn't contain an accidental coding standards violation.
PHPCS: ❌ ONLY KIDDING YOUR CODE IS NOT GOOD ENOUGH ❌

PR #286, opened 2022-07-28
Add stylelint tests

mattheu:
Add some basic tests for styles. Follows similar setup to eslint tests. Can increase coverage in future. 

@tomjn Already did a PR for updating to the latest config already: https://github.com/humanmade/coding-standards/pull/275. Some tests could be a nice way to help test this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,9 @@
 			}
 		]
 	},
+	"dependencies": {
+		"chalk": "^4.1.2"
+	},
 	"devDependencies": {
 		"@wordpress/npm-package-json-lint-config": "^4.0.5",
 		"lerna": "^5.5.0",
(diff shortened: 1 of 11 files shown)


mattheu:
Originally this included the update to the latest version of stylelint, but as there is already a PR for this, I figured I'd rework this to only include my tests.

package.json
```diff
@@ -67,6 +67,9 @@
 			}
 		]
 	},
+	"dependencies": {
+		"chalk": "^4.1.2"
```
tfrommen, line 71: Why do we need `chalk` as a (production?) dependency of the main package? 🤔

ntwb: Agreed @tfrommen, though as its the _root_ `package.json` this isn't much of an issue as this is never published

I'll also remove `chalk` from here in a follow up PR https://github.com/humanmade/coding-standards/blob/update-stylelint/packages/stylelint-config/package.json#L33-L35

There's a reasonable re-architecture of the packages coming shortly as I tweak everything for Lerna support that will _juggle_ many of the dependencies in the _root_ and package `devDependencies`

I'm ideally just getting this PR in a state that passes Travis to merge so I can begin on the above

ntwb: Actually, that can wait until the refactor, removed in e5ce448fa5f885daa3f595d12ec9c8166b0658e6

ntwb (approved):
Thanks a bunch @mattheu, apologies it took a while to land this

PR #287, opened 2022-07-28
Increase NPM version requirement and regenerate package lockfiles

mattheu:
[@tfrommen suggested in another PR that we regenerate the `package-lock.json` to use version 2 as a separate Pull Request to keep the size of the change down. ](https://github.com/humanmade/coding-standards/pull/275#discussion_r809226824)

One thing raised in that conversation was that we need to ensure backwards compatibility with Node 12 and NPM 6 as this is what is used by the Altis Build pipeline. [I'd just like to be clear that npm 6 does indeed support lockfile v2 so this change should be fine. ](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
       script:
         - vendor/bin/phpunit
     - language: node_js
-      node_js: 10
+      node_js: 16
       install:
         - npm install
         - cd packages/eslint-config-humanmade
@@ -41,16 +41,7 @@ jobs:
       script:
         - npm test
     - language: node_js
-      node_js: 12
-      install:
-        - npm install
-        - cd packages/eslint-config-humanmade
-        - npm install
-        - cd ../..
-      script:
-        - npm test
-    - language: node_js
-      node_js: 14
+      node_js: 18
       install:
         - npm install
         - cd packages/eslint-config-humanmade
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ jobs:
       install:
         - npm install
         - cd packages/eslint-config-humanmade
-        - npm install
+        - npm install --legacy-peer-deps
         - cd ../..
       script:
         - npm test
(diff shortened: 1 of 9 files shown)


tfrommen:
The issue I am (still) seeing with this is that people locally can use npm 6, and Node 12 or so. And everything is fine. Until they install or update a dependency, in which case they would generate a v1 lockfile, and again produce a massive change.

In the other PR, I said:

> if we decided to do require a modern verison of npm, I'd do this outside of this very PR.

By that, I did not mean to just use npm 7+ and push a v2 lockfile, but actually state that the coding standards (i.e., the ESLint and stylelint configs) require npm and Node in a specific version. As do not really require any specific version, but want people not to "downgrade" the lockfile, this should do:

```json
  "engines" : { 
    "npm" : ">=7.0.0",
    "node" : ">=15.0.0"
  }
```

mattheu:
But my understanding is we need this to run on node 12 and npm 6 (as is the current requirement of the Altis build pipeline unless it has changed since Feb. If so I don’t think we can set the engine above that, which could still cause generation of v1 lock files.

tfrommen:
> But my understanding is we need this to run on node 12 and npm 6 (as is the current requirement of the Altis build pipeline

Node 12 is what the build pipeline currently has installed, yes. But that is not a requirement in terms of "you cannot use anything newer". I know of projects that are using Node 14, and I think some are looking to use 16 or even Node 18.

mattheu:
@tfrommen I've updated this to add the following. 

* Specify requirement for Node 16 and npm 7. (I skipped node 15 as you suggested as 16 is the current LTS)
* Add documentation on using this in Altis build script.
* Remove the Travis CI tests for Node 10, 12 and 14. Replace with 16 and 18.

tfrommen (approved):
👍

ntwb (approved):
Looks good, thanks @mattheu

PR #288, opened 2022-08-03
Remove package.json lint rule for no devdeps

mattheu:
We are using `npm-package-json-lint` to check that our package.json files comply with certain rules. 

One of these rules is configured so that you are not allowed to use `devDependencies` in the packages, but it's OK at the top-level. However this test fails because we have devDependencies in the eslint config package. 

Seems to me that we should just disable this rule. I'm not sure why it has ever passed TBH.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
 					"requireEndingPeriod": true
 				}
 			],
-			"prefer-no-devDependencies": "warning",
 			"require-publishConfig": "error",
 			"require-repository-directory": "error",
 			"valid-values-author": [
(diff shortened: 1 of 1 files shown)


ntwb:
This is because the `devDependencies` need to be moved from the ESLint package to the root `package.json`

This just hasn't happened yet and is why this rule is only set to `warning` which is why CI runs currently pass

https://github.com/humanmade/coding-standards/blob/9aedd9d07d2575a694f9b13d0d1e0be410ae8274/package.json#L38

E.g. via https://app.travis-ci.com/github/humanmade/coding-standards/jobs/578616221#L262-L270

image

PR #307, opened 2023-02-03
Require double quotes for props in JSX

mattheu:
I recently had use inconsistent use of double/single quotes for props in JSX flagged in a pull request and was surprised this wasn't enforced in our coding standards. 

Thoughts on adding this?


rmccue:
That is surprising since I thought we did enforce that.

mattheu:
I did have to see if it was something we were inheriting from e.g. WP standards, but looks like it isn't 🤔

svandragt (approved):

PR #314, opened 2024-02-21
Upgrade wpcs

### LINKED ISSUE #313: Update to WPCS v3
We should track the latest version of the WP and a8c coding standards.

rmccue:
Fixes #313.

Have not exhaustively checked all the differences here yet to see what breaks.


rmccue:
Looks like the duplicate ignore checks may not be working as I think `::init` was removed, but it's not fatalling when testing locally with Real Projects™️ at least.

HM/Sniffs/Security/EscapeOutputSniff.php
```diff
@@ -14,6 +14,7 @@
  *
  * @see https://github.com/WordPress/WordPress-Coding-Standards/issues/1864
  */
+#[AllowDynamicProperties]
```
rmccue: @goldenapples Is this necessary? Seems like we should know the properties being used?

goldenapples: Huh. Yeah, this was pretty hacky.

The problem is that `$printingFunctions` is changed in the parent class from protected to private, so trying to unset items from `$this->printingFunctions` here is not going to work, and will trigger an accessing dynamic properties warning.

https://github.com/humanmade/coding-standards/blob/c5f69a8b9acaac86fe5f3c0854f3dd446a82c949/HM/Sniffs/Security/EscapeOutputSniff.php#L54-L56

This annotation silences the dynamic properties deprecation warning. It does not, however, make the code here work to ignore the safe printing functions.

tomjn:
@rmccue @goldenapples is there work still pending on this? It's unclear what the status of the PR is currently, from first glances it looks like Travis CI is the blocker

kadamwhite:
@tomjn I've been using this in production on a project via tracking the `dev-upgrade-wpcs` branch for several weeks with no issue. @rmccue @goldenapples I'd like to merge to get this landed and released, any objections? If not, I can forward with that next week.

composer.json
```diff
@@ -5,8 +5,8 @@
 	"license": "GPL-2.0-or-later",
 	"require": {
 		"php": ">=7.1",
```
kadamwhite: Should we bump the PHP requirement?
```suggestion
		"php": ">=8.0",
```

tomjn:
@kadamwhite I've been in a similar boat, from my POV the only blocker left for this is the failing Travis tests for a missing trait

tomjn:
I tried updating PHPBasicAuth with this branch and ran into the PHPCompatibility trim error. There's a v10 alpha that should fix this but given that it's an alpha would we even consider including that here? I know some projects have patches that work around the problem but fundamentally there are language syntax differences only v10 can account for

https://github.com/PHPCompatibility/PHPCompatibility/wiki/Upgrading-to-PHPCompatibility-10.0#updating-your-custom-ruleset-and-phpcs-ignore-annotations

kadamwhite:
@tomjn can you clarify how to reproduce the issue you're describing? I do not think we should be pinning an alpha, personally.

tomjn:
> @tomjn can you clarify how to reproduce the issue you're describing? I do not think we should be pinning an alpha, personally.

Likewise else I'd be adding commits, it seems that v10 coming out of alpha has a bit of a domino effect on downstream package releases. I forget where it was I was encountering this though

kadamwhite (approved):

rmccue:
🚀 🚀 🚀

Issue #23, opened 2017-09-18
Update wpcs to 0.13.1


connerbw:
"wp-coding-standards/wpcs": "^0.10.0"
https://github.com/humanmade/coding-standards/blob/master/composer.json#L7

Should be 0.13.1:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases

### Why?

Getting a lot of bogus errors when running Travis CI with PHP 7.2 (nightly)

```
Deprecated: Function create_function() is deprecated in /home/travis/build/[secure]/[secure]/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/WP/PreparedSQLSniff.php 

Warning: count(): Parameter must be an array or an object that implements Countable in /home/travis/build/[secure]/[secure]/vendor/wp-coding-standards/wpcs/WordPress/Sniffs/WP/I18nSniff.php on line 234

...
````

connerbw:
Related:

https://github.com/humanmade/coding-standards/pull/22

connerbw:
PHP 7.2 is planned to be released on November 30...

https://www.youtube.com/watch?v=f4zyjLyBp64

rmccue:
Sorted.

PR #333, opened 2026-05-15
Fix NamespaceDirectoryNameSniff misreading strict_types as namespace name

### LINKED ISSUE #319: NamespaceDirectoryNameSniff suggests `strict_types` in a subdirectory's file.
## Overview

When `strict_types` is declared in a subdirectory's namespaced file, `NamespaceDirectoryNameSniff` reports the `NameMismatch` error and recommends `strict_types` as the namespace and the directory name.

## Reproduction
Test file: `inc/subdir/class-test.php`
Test code:
```php
<?php

declare(strict_types=1);

namespace MyNamespace\Subdir;
```
Test command:
```php
./vendor/bin/phpcs --sniffs=HM.Files.NamespaceDirectoryName inc/subdir/class-test.php
```

Command result:
```
Directory subdir for namespace strict_types found; use strict_types instead
```

## Cause

`HM/Sniffs/Files/NamespaceDirectoryNameSniff.php` [Ref](https://github.com/humanmade/coding-standards/blob/01ccad39b28fd31b4e922c43e230b27a12966383/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php#L33)
Line: 33
```php
$name_ptr = $phpcsFile->findNext( T_STRING, 0);
```

`0` begins the search from the beginning of the file, not from where `T_NAMESPACE` was detected. The first `T_STRING` is `strict_types`.

## Workarounds

In PHP, the `strict_types` declaration must be the first statement in the script or a Fatal Error will be thrown, so this is not avoidable by rearranging line positions.

Therefore, the workarounds are:
- Don't namespace a file that declares `strict_types`, OR
- Don't declare `strict_types` in a namespaced file.

## Solution
The solution is straightforward:

Change `0` to `$stackPtr` so the search begins from where `T_NAMESPACE` was detected:
```php
$name_ptr = $phpcsFile->findNext( T_STRING, $stackPtr);
```

ewartx:
## Summary

Fixes #319.

`NamespaceDirectoryNameSniff` calls `findNext( T_STRING, 0 )` on line 33 to locate the namespace name, which starts searching from position 0 of the file. When the file opens with a `declare(strict_types=1);` statement before its namespace declaration, the first `T_STRING` token in the file is `strict_types` (PHP's tokenizer emits the declare directive name as `T_STRING`), so the sniff reads `strict_types` as the namespace and reports a false `NameMismatch` or `ExtraDirs` error against the actual directory.

## Fix

Change the `findNext` start position from `0` to `$stackPtr`, so the search begins at the `T_NAMESPACE` token the sniff is currently processing rather than the beginning of the file.

```diff
-		$name_ptr = $phpcsFile->findNext( T_STRING, 0);
+		$name_ptr = $phpcsFile->findNext( T_STRING, $stackPtr );
```

This matches the resolution proposed in #319 directly.

## Why this is safe

The fix produces the same result as the existing code in every case except the buggy one:

- **Plain `<?php namespace Foo;`** (no `declare`): the first `T_STRING` in the file is the namespace name itself, so the original (`from position 0`) and the fix (`from $stackPtr`) land on the same token.
- **`<?php declare(strict_types=1); namespace Foo;`**: original reads `strict_types`; fix correctly reads `Foo`.
- **Multi-namespace `namespace A {} namespace B {}` files**: original reads `A` for both invocations (the file-first `T_STRING`); fix correctly reads each block's own namespace name.

Only `declare()` directives can legally place a `T_STRING` token before `T_NAMESPACE`; PHP requires `namespace` to be the first non-declare statement, so other cases aren't reachable in real code.

## Test

Adds `HM/Tests/Files/NamespaceDirectoryNameUnitTest/inc/standards/strict-types-namespace.php` to the existing fixture tree and adds its basename to the `$pass` list in `NamespaceDirectoryNameUnitTest::getErrorList()`. The fixture is the minimal repro from the issue:

```php
<?php

declare(strict_types=1);

namespace HM\Coding\Standards\StrictTypesNamespace;
```

Verified locally with `./vendor/bin/phpunit --filter NamespaceDirectoryNameUnitTest`:

- Without the sniff fix, the new fixture fails: `Expected 0 error(s) in strict-types-namespace.php but found 1 error(s). Directory /standards for namespace strict_types found; nested too deep. (HM.Files.NamespaceDirectoryName.ExtraDirs)`
- With the sniff fix, the full `NamespaceDirectoryNameUnitTest` passes.

### DIFF
--- a/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
+++ b/HM/Sniffs/Files/NamespaceDirectoryNameSniff.php
@@ -30,7 +30,7 @@ class NamespaceDirectoryNameSniff implements Sniff {
 		$tokens = $phpcsFile->getTokens();
 		$namespace = '';
 
-		$name_ptr = $phpcsFile->findNext( T_STRING, 0);
+		$name_ptr = $phpcsFile->findNext( T_STRING, $stackPtr );
 		if ( ! $name_ptr ) {
 			// Non-namespaced, skip check.
 			return;
--- a/HM/Tests/Files/NamespaceDirectoryNameUnitTest.php
+++ b/HM/Tests/Files/NamespaceDirectoryNameUnitTest.php
@@ -51,6 +51,7 @@ class NamespaceDirectoryNameUnitTest extends AbstractSniffUnitTest {
 			'namespace.php',
 			'camelcased-namespace.php',
 			'underscored-namespace.php',
+			'strict-types-namespace.php',
 		];
 		if ( in_array( $file, $pass, true ) ) {
 			return [];
--- a/HM/Tests/Files/NamespaceDirectoryNameUnitTest/inc/standards/strict-types-namespace.php
+++ b/HM/Tests/Files/NamespaceDirectoryNameUnitTest/inc/standards/strict-types-namespace.php
@@ -1,0 +1,5 @@
+<?php
+
+declare(strict_types=1);
+
+namespace HM\Coding\Standards\StrictTypesNamespace;


ewartx:
@kadamwhite tagging you for visibility — this addresses the upstream NamespaceDirectoryNameSniff bug discussed in the PR #23 thread on our project repo. Fix matches the resolution proposed in #319; local test run confirms it. Happy to iterate on anything.

kadamwhite (approved):
Confirmed this resolves the local issue and the fixture does not pass without the change.

PR #335, opened 2026-05-28
Relax WPCS version constraint to ^3

kadamwhite:
PHPCS will be pulled in as a transient dependency of both WPCS and VIPCS, so keeping it in our own composer.json does nothing but increase our maintenance burden.

The progressive point updates to WPCS in 3.1, 3.2 and 3.3 have improvements which would benefit HM projects and which make the coding standards more intuitively support modern versions of WordPress.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,11 +5,10 @@
 	"license": "GPL-2.0-or-later",
 	"require": {
 		"php": ">=8.2",
-		"wp-coding-standards/wpcs": "~3.0.0",
-		"automattic/vipwpcs": "~3.0.0",
+		"wp-coding-standards/wpcs": "^3.0.0",
+		"automattic/vipwpcs": "^3.0.0",
 		"fig-r/psr2r-sniffer": "^0.5.0",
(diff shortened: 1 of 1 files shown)


joehoyle:
@kadamwhite given we integrated directly with PHP_Codesniffer, it would make sense to have it as a dependency. E.g. https://github.com/humanmade/coding-standards/blob/main/HM/Sniffs/Classes/OnlyClassInFileSniff.php. The reason we specify it is given we have code that relies specifically on certain versions of PHP Codesniffer means we should apply our own version constraint.

kadamwhite:
@joehoyle what would you recommend for a constraint, WPCS specifies ^3.13.4

joehoyle:
@kadamwhite it depends what we _are_ compatible with, which I dont know. Perhaps it safe to go `^3`, but I'm not totally sure!

kadamwhite:
@joehoyle Reviewed our custom changes, ^3 feels safe. Since WPCS requires ^3.13 I've gone with that.

joehoyle (approved):

PR #337, opened 2026-06-22
Permit single-file mu-plugins

kadamwhite:
I reviewed use of HMCS across a variety of VIP and Altis projects and noticed a recurring configuration in project PHPCS to disable the rules which prevent the use of single-file MU plugins, e.g. to drop a small `hm-thirdpartyplugin-integration.php` which simply registers a few hooks to customize third-party code and doesn't benefit practically from the more robust folder-based code structure.

I believe this should be permitted by our standards, and this PR "paves the cowpath" by adjusting the ruleset to avoid false-flagging files like this from needing to be organized into `inc/namespaces`, and to permit them to have side-effects.

### DIFF
--- a/HM/Sniffs/Files/MuPluginFileTrait.php
+++ b/HM/Sniffs/Files/MuPluginFileTrait.php
@@ -1,0 +1,27 @@
+<?php
+namespace HM\Sniffs\Files;
+
+/**
+ * Shared helper for detecting single-file must-use plugins.
+ *
+ * Single-file mu-plugins live as direct children of a `mu-plugins/` directory
+ * (or `client-mu-plugins/`, on VIP). This is a useful pattern to support, but
+ * by definition can't be split into an inc/ directory or named namespace.php
+ * (multiple plugins would collide), and should allow side effects.
+ */
+trait MuPluginFileTrait {
+	/**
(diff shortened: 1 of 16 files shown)


HM/Sniffs/Files/MuPluginFileTrait.php
```diff
@@ -0,0 +1,27 @@
...
+	 * @param string $path Full path to the file being checked.
+	 * @return bool True if the file is a direct child of mu-plugins/ or client-mu-plugins/.
+	 */
+	protected function is_single_file_mu_plugin( $path ) {
+		// Normalize the directory separator across operating systems.
+		if ( DIRECTORY_SEPARATOR !== '/' ) {
+			$path = str_replace( DIRECTORY_SEPARATOR, '/', $path );
+		}
+
+		return (bool) preg_match( '#/(?:client-)?mu-plugins/[^/]+\.php$#', $path );
```
joehoyle, line 25: Is `plugins-mu` also used?

kadamwhite: Not in any projects that I've worked on in the past 3–4 years, my inclination is to leave that one out as it's a confusing pattern

joehoyle (approved):

PR #340, opened 2026-06-22
Fix false-positive escaping warning on get_block_wrapper_attributes

kadamwhite:
[get_block_wrapper_attributes](https://github.com/WordPress/WordPress/blob/7.0-branch/wp-includes/class-wp-block-supports.php#L166-L240) responsibly builds its output string using safecss_filter_attr and esc_attr, and its output is not filterable. We should allow this function.

(This is particularly annoying to get false-flagged today, because it usually occurs within an inline <?php ... ?> tag in HTML partials that makes it visually clunky to add an ignore directive.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Next Patch
+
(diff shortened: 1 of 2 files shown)


rmccue (approved):
👍 Looks right based on the core implementation - which itself doesn't actually echo anyway.

PR #341, opened 2026-06-23
Update changelog for 2.2.1

kadamwhite:
Also attempts to propose proactively filling out prerelease version changes in the changelog, so we can merge and tag rather than needing these changelog release update PRs. Thoughts?


rmccue (approved):
:laff:

PR #342, opened 2026-06-26
Exclude built code folders from stylistic PHPCS rules

kadamwhite:
Built asset files are standard under current wp-scripts best practices, so our ruleset should not enforce human-authorship-oriented sniffs on these autogenerated files.

I've limited our exclusions to the default `build/` and oft-used `dist/` folder prefixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and fill in the contents as you go. This simplifies later release management. --
 
 ## 2.3.0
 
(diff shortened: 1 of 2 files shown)


rmccue (approved):
I think this makes sense; I assume because this isn't prefixed with `*` that eg `inc/tools/build/foo.php` would still be included?

kadamwhite:
> I assume because this isn't prefixed with * that eg inc/tools/build/foo.php would still be included?

Correct!

PR #347, opened 2026-07-07
Introduce sniff to avoid doing_it_wrong usage of register_block_type

kadamwhite:
`register_block_type` internally delegates to `register_block_type_from_metadata` if a readable file path is passed, but if the file does not exist (usually from not having run a build) it throws a confusing `_doing_it_wrong` message.

We should be using `register_block_type_from_metadata` directly whenever we are not passing the name of a block as the first argument, since this can lead to wasted debugging time if the error is not understood correctly.

This sniff attempts to put in place some heuristic detection for path-y first arguments passed to `register_block_type`, to catch this in PRs / CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 <!-- When submitting a PR for a feature, add the appropriate prerelease heading here
 and fill in the contents as you go. This simplifies later release management. -->
 
+## 2.4.0
+
+- Add `HM.Functions.RegisterBlockTypePath` sniff recommending `register_block_type_from_metadata()` when a file path is passed to `register_block_type()`
+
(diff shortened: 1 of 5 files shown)


kadamwhite:
Tagging a few of you for feedback on whether this feels worth adding to our standard. It's something that's been a slight footgun even after trying to add agent instructions to support using the metadata version, and I'd like to be able to flag it at the PR stage automatically, without wasting human cycles on it — but it's technically already covered by a `_doing_it_wrong`, assuming the developer is monitoring logs for those properly.

roborourke (approved):
It's a simple fix and if it avoids a common problem then it gets my vote.

